PR #9, opened 2015-10-05
Miscellaneous fixes

radioactiveman:
Two more fixes from [Audacious](https://github.com/audacious-media-player/audacious-plugins/commits/master/src/adplug):
* Fix incorrect format strings
* Update FSF address


Malvineous:
Thanks for the updates!

PR #13, opened 2016-03-21
Fix loudness music files always playing at 70hz

illydan:
Loudness Sound System files are fixed to be replayed at 70hz but there are files which need to be replayed at 60hz.
This uses the replay rate stored inside the file instead of forcing 70hz


Malvineous:
Thanks very much for the patch!

PR #18, opened 2016-06-02
ROL: Fixes for pitch and volume event handling and code refactoring. …

OPLx:
This fixes some latent bugs from back when I had originally implemented the ROL player specifically with pitch event handling (was completely broken), volume events (the instrument's carrier _**total level**_ value was not properly preserved), and writing past register 0xC8 (Feedback / Algorithm) when playing files that are in percussive mode.  The test for ROL files was also updated to reflect the change in register writes as well.

Also done was some code refactoring to clean things up a bit and resolve the warnings that the compiler found; I compiled and tested only with Visual Studio 2015.

Hopefully these changes will be useful for AdPlug! :)


Malvineous:
Merged in 85dde0a, many thanks!

PR #19, opened 2016-07-03
ROL: Display used instrument names

binarymaster:
This will allow to see used instruments in the ROL file

### DIFF
--- a/src/rol.cpp
+++ b/src/rol.cpp
@@ -710,18 +710,9 @@ void CrolPlayer::load_instrument_events(binistream *f, CVoiceData & voice,
         event.time = static_cast<int16_t>(f->readInt(2));
         f->readString(event.name, ROL_MAX_NAME_SIZE);
 
-        for (uint8_t j = 0; j < ROL_MAX_USED_INSTRUMENTS; j++)
-        {
-            if (!*instname[j])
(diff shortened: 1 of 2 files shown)


Malvineous:
Many thanks for the patch!  This one looks fine, but would you mind changing it to avoid using C-style chars?  It's a bit of a mix of C and C++ at the moment, so it would be nice if it could be all C++.  I know you said you weren't an expert in C++, so here are some hints that I am thinking of:

* Use `std::vector<std::string>` instead of `char instname[][]`.  A vector is like a resizable array of string objects.
* Use the vector's [`push_back()`](http://en.cppreference.com/w/cpp/container/vector/push_back) method to add the string instead of `strcpy`
* If `instname` is an `std::vector` then you can still use `instname[2]` to get the third instrument name in `getinstrument()`.  Just use the vector's [`size()`](http://en.cppreference.com/w/cpp/container/vector/size) method first to check if the value is in range or not.

If you are willing to make these changes then it will keep the code in the C++ style and I'll be happy to merge them in!

binarymaster:
Done! It looks better now :)

Malvineous:
Perfect!  Many thanks.  Merged in 28561a8.

PR #21, opened 2016-07-20
CRealopl support in Linux 

tomas:
This PR allows talking to an OPL chip directly in Linux via the CRealopl class. 

Given that we need to ask for permission before sending any commands, I added a `gotperms` flag to check whether we've already done so or not, so we only request `ioperms` once.

I also noticed that the CrawPlayer wouldn't produce any sound without first setting all registers to zero, [as the DRO player does](https://github.com/adplug/adplug/blob/master/src/dro.cpp#L144-L151) so I also added that logic as part of CRealopl's init logic -- didn't want to fiddle with the players. :)

Finally, I also cleaned up and reformatted part of the code. I didn't see any pattern with regards to code style so I hope this is OK.


Malvineous:
Merged, thanks!

tomas:
Woohoo!

PR #22, opened 2016-08-12
Add support for GOT format

binarymaster:
This pull request adds support for God of Thunder Music format.

Tested and verified playback and timings using this specification:
http://www.shikadi.net/moddingwiki/God_of_Thunder_Music_Format


src/got.cpp
```diff
@@ -0,0 +1,115 @@
...
+	data = new Sdata[size];
+	for(unsigned int i = 0; i < size; i++) {
+		data[i].time = f->readInt(1);
+		data[i].reg = f->readInt(1); data[i].val = f->readInt(1);
+	}
+
+	CAdPlugDatabase::CKey menu_music;
+	menu_music.crc16 = 46631;
+	menu_music.crc32 = 1912826945;
+
```
binarymaster: One of the tests fails with gcc/clang, there's some problem with checksums. I'm using CRC16/32 with `CAdPlugDatabase::CKey` to detect **Menu Song** file which has different playback frequency.

Build using VS2015 succeeds (as well as AppVeyor).

Any ideas why this is happening?

Malvineous:
I wonder whether the CRC calculation functions are buggy?  Could be a 32-bit vs 64-bit problem?  Do you think, as a separate pull request, you could write some tests for the CRC code?  You should be able to duplicate `emutest.cpp` and change it to calculate some CRC values then check them against hard-coded constants.  Maybe you could CRC a few of the test files.

If your tests work for AppVeyor but fail for Travis then that will be pretty clear the CRC code needs adjustment, and then we'll have CRC tests to make sure this isn't a problem again one day.

binarymaster:
Previously I obtained CRC values by debugging the variables.

Updated values from WinHex - I think it's a reliable source.

> Could be a 32-bit vs 64-bit problem?

Yes, according to the [tests](https://github.com/adplug/adplug/pull/23), this problem also exists.

Malvineous:
CRC tests look good now!  Merged.

binarymaster:
Thanks! Also see https://github.com/adplug/winamp/pull/3 for Winamp plugin.

PR #23, opened 2016-08-13
CRC tests for AdPlug

binarymaster:
Actual CRC16/32 constants are obtained using WinHex.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -55,7 +55,7 @@ build_script:
 
     appveyor-retry nuget update contrib\vs2015\vs2015.sln
 
-    FOR %%T IN (v140,v120_xp) DO ( FOR %%P IN (x86,x64) DO ( FOR %%C IN (Debug,Release) DO ( FOR %%F IN (adplug,emutest,playertest) DO ( echo *** Building %%F as %%T/%%P/%%C *** && msbuild contrib\vs2015\%%F\%%F.vcxproj /p:Configuration=%%C /p:Platform=%%P /p:PlatformToolset=%%T /p:SolutionDir=..\ /v:minimal /nologo || EXIT 1 ) ) ) )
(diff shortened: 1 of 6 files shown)


binarymaster:
So... we have some problems in CRC code.

```
AppVeyor output:
Checking CRC: 2001.MKJ [FAIL: 20792, 4198819144]
Checking CRC: ADAGIO.DFM [FAIL: 33, 4230051849]
Checking CRC: adlibsp.s3m [FAIL: 16552, 3282359743]

Linux GCC output:
Checking CRC: 2001.MKJ [FAIL: 20792, 18446744072657415077]
Checking CRC: ADAGIO.DFM [FAIL: 33, 18446744071446523163]
Checking CRC: adlibsp.s3m [FAIL: 16552, 18446744071448425428]
```

And all values differ from constants.

Malvineous:
Very nice work!  Job well done.  Many thanks!

Before I merge it, could I ask for a couple of minor changes?  Could you split the CRC16/32 check so that if the CRC16 passes but the CRC32 fails (or vice versa), that is shown in the output?  Currently if one passes but the other fails, it's not clear which one has the problem.  Having two lines of output per file would be fine (e.g. "Checking CRC16" and "Checking CRC32").

Also it would be nice to add `std::hex` to the `std::cout` line so the failure values are in hex to match the hard-coded constants.

How did you calculate the hard-coded constants?  Did you use an external CRC algorithm or did you use the AdPlug code to calculate them?

Thanks again!  This is great.

Malvineous:
Sorry just realised what you meant by WinHex.  I'm not sure whether that CRC algorithm is the same as the AdPlug one.

I think we'd need to compile AdPlug for 32-bit and then use those figures as the desired CRC output.  The AdPlug database already uses those, so they are really the values we want to obtain on all platforms.

The current AppVeyor test that fails is for 32-bit (see [line 165](https://ci.appveyor.com/project/AdPlug/adplug/build/2.3-beta39#L165) - it says `x86`) so I think those should be the constants we are going for.  These may not be the "correct" CRC ones, but they avoid the need to regenerate the AdPlug database.

binarymaster:
Oh, so AdPlug uses its own CRC algorithms, which are not compatible with classic CRC16/32.

Allright then, updating the code... :)

Malvineous:
Looks like it!  Thanks to your tests I have figured out the bug in the AdPlug CRC code, and fixing this makes your tests pass on a 64-bit machine, when using the constants from AppVeyor 32-bit.

So if you can update this I will merge your tests and then merge the fix, which should fix the error in the GoT PR.

binarymaster:
Sounds good! I corrected hashes and splitted test.

Malvineous:
Perfect!  Merged.  Thanks again :-)

PR #28, opened 2016-09-04
adtrack: Close instrument file (fix #25)

### LINKED ISSUE #25: File descriptor leak in CadtrackLoader::load()
At adtrack.cpp line 65 there is a check that the instruments file (instf) is exactly 468 bytes.  If not, CadtrackLoader::load() returns false without closing the file.

binarymaster:


Malvineous:
Committed in 65a5c82, thanks again!

PR #29, opened 2016-09-04
ADL: Fix out of index read (fix #27)

### LINKED ISSUE #27: Reading past end of AdlibDriver::_regOffset array
AdlibDriver::_regOffset[] is an array of 9 bytes (see adl.cpp line 2048).  However, the first iteration of the _curChannel loop in AdlibDriver::executePrograms() reads from _regOffset[9], which is past the end of the array.

binarymaster:


Malvineous:
Committed in 4e98d16, thanks!

PR #30, opened 2016-09-04
XAD/BMF: Fix unterminated strings (fix #26)

### LINKED ISSUE #26: Unterminated string copy in CxadbmfPlayer
CxadbmfPlayer::xadplayer_load() uses strncpy() to initialize 36-byte "title" and "author" strings in the "bmf" struct but does not ensure that they are null-terminated.  Then xadplayer_gettitle() and xadplayer_getauthor() copy the string data with no length limit, which could result in reading past the end of the array.

binarymaster:


Malvineous:
Committed in facbeae

PR #35, opened 2016-09-16
add Nuked OPL3 emulator

loki666:
add Nuked OPL3 v1.7.4 from @nukeykt
the files opl3.c and opl3.h are from chocolate-doom with GPL License, but @nukeykt is ok to switch to LGPL.


binarymaster:
I've heard of Nuked OPL3 before, it's pretty good.

However there are only added files in your PR, but I don't see any integration attempts that will allow to use it for playback.

Are you planning to do so?

loki666:
what do you mean?
add it to Makefile.am ?

binarymaster:
Yes, at least. See this commit about WoodyOPL for the reference:

https://github.com/adplug/adplug/commit/f564c8c0e9276183ec5e32b688a67624bd57ec42

And also this:

https://github.com/adplug/adplug/search?utf8=%E2%9C%93&q=wemuopl.h

loki666:
done, do you guys need an approval from @nukeykt to change the license ?

Malvineous:
It would be good to get something in writing to confirm LGPL is OK for Nuked.

I'll leave the PR unmerged until someone contributes code to make it usable with AdPlug (and we get approval for LGPL.)  It would also be good to get a PR for adplay-unix so the new OPL synth can be tested from the command line as well. @loki666 please advise whether you are willing to do this or whether you would need someone else here to assist.

loki666:
@Malvineous I'll take a blind shot at adplay-unix, and ask @nukeytk to comment here

nukeykt:
Hi!
I'm author of the Nuked OPL3 emulator and you can use this version under LGPL 2.1 :)

loki666:
I made a PR for adplay-unix

Malvineous:
@nukeykt Many thanks for your consent!

src/opl3.h
```diff
@@ -0,0 +1,135 @@
...
+//          Tremolo and phase generator calculation information.
+//      OPLx decapsulated(Matthew Gambrell, Olli Niemitalo):
+//          OPL2 ROMs.
+//
+// version: 1.7.4
+//
+
+#ifndef OPL_OPL3_H
+#define OPL_OPL3_H
+
```
Malvineous: I would suggest renaming the files from `opl3.c` and `opl3.h` to `nukedopl.c` and `nukedopl.h` so it's clear what they are.  The existing AdPlug OPL synth files are somewhat poorly named so let's not continue that tradition!  This would involve updating this `#define` constant too for consistency, however you could leave the `#define` as is if you split out the `class` definition of `CNemuopl` into a separate `.cpp` file as suggested above.  The reason is that if you split out the class definition, then only one `.cpp` file will ever `#include` this file, so the `#define` value is not particularly important in that arrangement.

src/nemuopl.h
```diff
@@ -0,0 +1,56 @@
...
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * nemuopl.h - Emulated OPL using the Nuked OPL3 emulator
+ */
+
+#ifndef H_ADPLUG_NEMUOPL
+#define H_ADPLUG_NEMUOPL
+
+#include "opl.h"
+extern "C" {
```
Malvineous: Is `extern "C"` needed here?  I don't think we are linking with an external C library for the OPL stuff.

src/nemuopl.h
```diff
@@ -0,0 +1,56 @@
...
+#include "opl3.h"
+}
+
+class CNemuopl: public Copl
```
Malvineous: I would move this class definition into `nemuopl.cpp` (along with the `#include` line for Nuked OPL) so that code using this OPL synth doesn't have to include the whole Nuked OPL `.h` file, which could cause identifier conflicts.  Doing this would avoid the need to encapsulate the Nuked OPL `.h` file inside a `namespace` too.

I realise the existing code does not do this, but it would be nice to make improvements here while we can.

src/nemuopl.h
loki666: It is, nukedopl.c is a C module

Malvineous: `nukedopl.c` is source file though, you only need `extern` when you're linking with a library written in C (e.g. a `.dll` or `.so` file.)  So should be safe to remove.  You should find when you take it out, everything compiles fine and you can still run `adplay-unix` correctly when set to Nuked OPL.

loki666: if I remove the extern "C" i get

`
../../src/nemuopl.cpp:40: error: undefined reference to 'OPL3_GenerateStream(_opl3_chip*, short*, unsigned int)'
../../src/nemuopl.cpp:45: error: undefined reference to 'OPL3_WriteRegBuffered(_opl3_chip*, unsigned short, unsigned char)'
../../src/nemuopl.cpp:29: error: undefined reference to 'OPL3_Reset(_opl3_chip*, unsigned int)'
`

Malvineous: Oh you are right, my apologies.  I have always renamed `.c` files to `.cpp` in my projects, even if the code is in C, but apparently leaving the `.c` extension effectively adds `extern C` over the `.c` file, so you do need to add the `extern` line when using it from a `.cpp` file.

Thank you for testing this!

Malvineous:
Merged in 14e57b84c2c8d2489a1f1b8e26376a4f251f7394, many thanks!

nukeykt:
You forgot to change GPL to LGPL in nukedopl.c/h

Malvineous:
@loki666 Are you able to supply a patch to update the licences?

Malvineous:
License updated in #37.

PR #42, opened 2016-10-16
MUS/IMS/MDI: Add new formats

### LINKED ISSUE #36: Add support for ROL-derived formats (MUS, IMS, MDI)
These formats were popular in many DOS games. They all are based on AdLib Sound Driver by Ad Lib Inc. Curious that they weren't added to AdPlug before.

Now I'm trying to port it into AdPlug from old DOS sources, especially `ADLIB.C` which dated back to 1989. Here is my branch with first implementation:

https://github.com/binarymaster/adplug/tree/adlib-formats

It successfully compiles in VS2015, but the player sounds wrong. I'm pretty sure that replaying code is correct (maybe except timing), but the problem is with note generation code - functions `NoteOn`, `SetFreq` and possibly others.

Well, it would be good to get some assistance on this. I think @OPLx could help with figuring out the problems in the code, since he wrote ROL replayer, which also uses `ADLIB.C` functions. If we fix all playback issues, AdPlug would get support for 3 old-new formats!

I attached the source code of players, music for testing in different formats + raw captured output of some songs. Hope the AdPlug developer community will help. :smile: 

Attachment: [adlib-music.zip](https://github.com/adplug/adplug/files/483775/adlib-music_zip.txt)

binarymaster:
This PR fixes #36.

Ready to merge!


binarymaster:
Some problems with arrays... however this builds successfully in VS2015 with Debug and Release configurations.

@Malvineous any ideas?

binarymaster:
Now builds successfully in AppVeyor.

Travis compiles, but shows strange error:
```
/bin/bash: line 5: 12937 Segmentation fault      (core dumped) ${dir}$tst
```

Related to `playertest`.

PR #44, opened 2016-11-04
CMF: Fix duplicated note and effect issues

### LINKED ISSUE #41: Missing CMF Notes
There are notes missing on CMF playback. I think it could be to do with the number of instruments (and so could potentially be an easy fix, I hope). @Malvineous, what do you think?

Attached is an example of a file from the game Xargon. Included are the CMF file and two different recordings of it (one through ADPlug and one converted with a CMF Tool several years back). You'll notice that the bass notes are consistent throughout the CMF Tool's output, whereas ADPlug skips over some of them (leaving spaces).

Here's an example: [example.zip](https://github.com/adplug/adplug/files/529125/example.zip)

binarymaster:


Malvineous:
Hmm I'm a little unsure about whether this is the correct solution to the problem.  It seems odd that playing a note when one is already playing should silence it, then when the note-off is played the note should sound again.

It seems like there's more to it, like perhaps the note-on event and the note-off event are being processed in the wrong order when they occur at the same instant.

Any thoughts?

binarymaster:
It's illogical to process the sequence of events in non-sequential manner, and that would be more complex to program. Rather there can be some sort of MIDI-to-OPL channel mapper which behaves like that.

To know how exactly the events are being processed, we'll need to decompile SBFMDRV. But I just suggested how this mapper works, and made a quick fix. Actual algorithm can be more complex, but I didn't dig into it. So this patch requires testing on different music files.

binarymaster:
TODO: dynamic note frequency changes for transpose controller and pitch bends.

binarymaster:
I've added a fix for transpose effect:
- Note transposition is related to a MIDI channel (pretty same as Pitch Bend), not a whole song
- Division by an integer (`this->iTranspose / 128`) was incorrect, changed to float
- Note frequencies are updated in realtime now (when event occured)
- Division by 128 is not enough, changed to 256 to match SBFMDRV output
- Added test music file which uses transpose effects

Malvineous:
I'll definitely have to look into this to confirm exactly what SBFMDRV is doing.  I have a particular interest in the CMF format so I want to understand exactly what's going on there.

With the transposition, how did you work out that you need to divide by 256 instead of 128?  All the documentation I can find says you should divide by 128, and I thought I had tested this with SBFMDRV as well, so if this is incorrect I would definitely like to understand why!

binarymaster:
> With the transposition, how did you work out that you need to divide by 256 instead of 128?

When I implemented real-time note frequency changes, I tested AdPlug playback with `1.CMF` song (see [this attachment](https://github.com/adplug/adplug/files/598669/CMF_zip.txt)), expecially vibrato effect at 0:52 that achieved by the sequence of transpose control changes.

Then I tested same music file with PlayCMF (using SBFMDRV) in DOSBox. The vibrato effect has lower expression than in AdPlug. The divider 256 was found by trial and error method and an ear of music.

I think this also can be tested (confirmed or not) by logging FNum values produced by SBFMDRV and AdPlug while playing that vibrato sequence.

Malvineous:
@binarymaster I guess you can probably merge this if you're happy it solves the problem, but if you can leave the matching issue (#41) open as a reminder for me to investigate further that would be good!  I'm happy if this fixes the problem but I'm convinced there is another explanation and would like to follow up when I get a chance, so hopefully the open issue will remind me.

binarymaster:
Ok! I'll merge only my first commit for missing notes, because the second for note transposing is breaking the tests (I'll try to fix it later).

PR #56, opened 2017-11-14
More fixes

SoapGentoo:
* Do not use `AM_MAINTAINER_MODE`
  `AM_MAINTAINER_MODE` has been considered a massive
  design flaw and leads to subtle bugs. For instance
  building Adplug as a live version under Gentoo
  does not enable all of the required bootstrap rules,
  leading to idiosyncratic build failures.
  See also:
    https://www.gnu.org/software/automake/manual/html_node/maintainer_002dmode.html

* Do not assign string literals to `char*` vars:
```
adplugdb.cpp:113:1: warning: ISO C++ forbids converting
    a string constant to ‘char*’ [-Wwrite-strings]
     };
     ^
```

### DIFF
--- a/test/Makefile.am
+++ b/test/Makefile.am
@@ -6,8 +6,7 @@ emutest_SOURCES = emutest.cpp
 
 crctest_SOURCES = crctest.cpp
 
-AM_LDFLAGS = $(top_builddir)/src/.libs/libadplug.la $(libbinio_LIBS)
-
+LDADD = $(top_builddir)/src/.libs/libadplug.la $(libbinio_LIBS)
 AM_CPPFLAGS = $(libbinio_CFLAGS)
 
 TESTS = playertest emutest crctest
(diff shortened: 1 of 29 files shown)


Malvineous:
@SoapGentoo First two commits look good, but for the Autotools stuff would you mind also updating the `README` (since it mentions maintainer mode) and also confirming that the changes both 1) allow building with `doc/version.texi` present, and 2) rebuild `doc/version.texi` as needed (e.g. if it's deleted.)  If that works then I'll merge in the changes.  Thanks!

SoapGentoo:
@Malvineous yes, double-checked it (that is specifically why I removed `AM_MAINTAINER_MODE `). Now, the rules producing `doc/version.texi` are **always** active, such that you should never have a stale tree. Also removed the mention of `--enable-maintainer-mode` in the `README`. Lastly, instead of running `make check`, I changed Travis to run `make distcheck`. This has the advantage of always flexing the out-of-source/VPATH builds, which is important nowadays, as touching the source-tree is a big no-no.

I have also added some small additional fixes to fix some more niggles:

1. two test files (`MainBGM5.vgm` and `MainBGM5.ref`) weren't being added to the tarball
2. `test/crctest.cpp` would not run when building out-of-source (due to never using the value of `srcdir`)

test/Makefile.am
```diff
@@ -33,4 +32,4 @@ EXTRA_DIST = 2001.MKJ 2001.ref ADAGIO.DFM ADAGIO.ref adlibsp.ref adlibsp.s3m \
 	go-_-go.ref Flying.mdi Flying.ref RIK6.MDI RIK6.ref NECRONOM.CMF \
 	NECRONOM.ref YsBattle.vgm YsBattle.ref BeyondSN.vgm BeyondSN.ref \
 	crusader.raw crusader.ref GALWAY.SOP GALWAY.ref bmf1_1.bmf bmf1_1.ref \
-	2.CMF 2.ref
+	2.CMF 2.ref MainBGM5.vgm MainBGM5.ref
```
binarymaster: Good catch! Indeed, I forgot to add these files to Makefile.

SoapGentoo:
@Malvineous it's ready now with some additional diagnostic printouts in the configure script.

Malvineous:
Looks good, many thanks for the fixes!

PR #59, opened 2017-11-20
Make build system completely non-recursive

SoapGentoo:
@Malvineous as promised. Writing the required files one per-line make it easier to insert and sort everything alphabetically. I think the remaining root Makefile.am is pretty readable, given how I structured it.


Malvineous:
This looks really nice!  I only have one reservation.

The test binaries are still built inside the `test` folder (great) but if you're in the test folder you can no longer run them (`cd test && ./playertest`) because they are now loading the data files from the `./test/` subdirectory.  This means you have to go out of the `test` subdir in order to run the tests, even though the binaries and data are inside that folder.

I can't really think of a better way of doing this (I take it you can't `cd` into the folder in a subshell when running each test?), so maybe it's something I will just have to get used to.  When you're in the `test` folder you can't run `make` anymore either, so maybe the way you have it is the better way as you shouldn't be working inside any of the subfolders any longer.

SoapGentoo:
@Malvineous by using `AM_TESTS_ENVIRONMENT` I could restore the old behaviour, so I guess your last reservation has been solved :wink:

Malvineous:
Nice job, that's perfect!  Many thanks!

PR #62, opened 2018-01-10
Optimize of Nuked OPL3 emulator

troosh:
It would be great if someone did code review and benchmarking of my patches to optimize the Nuked OPL3 emulator from @nukeykt (I don't know the homepage of this project, maybe https://github.com/chocolate-doom/chocolate-doom/blob/master/opl/opl3.c is better).

These patches are for one marginal CPU architecture (not for x86, ARM and other well know), which means I can not guarantee that they will all improve performance and work correctly. In general, more research is needed, for which I did not have the time and desire.

P.S. Should not expect a significant speedup: on out-of-order execution CPUs, will be about 5%. And total size of the library will decrease by 4 KB. However, this may open the way for further optimizations.


src/nukedopl.c
```diff
@@ -556,6 +556,12 @@ static void OPL3_SlotWriteE0(opl3_slot *slot, Bit8u data)
     case 5:
         slot->phaseshift = 1;
         break;
+    case 6:
+        slot->phaseshift = 16; // set phase to zero and flag for non-sin wave
+        break;
+    case 7:
+        slot->phaseshift = 32; // no shift (work by mod 32), but flag for non-sin wave
```
troosh: This place may not be portable on some architectures.

Malvineous:
@nukeykt: This is probably one for you!

nukeykt:
Thanks. Looks good. BTW, what do you think about approach used in MAME core, i.e. using precalculated waveforms LUTs?

loki666:
@troosh what are the performances on a Cortex-M4 ?

troosh:
I have not benchmarked this yet. But you can do it yourself. It was about large tables - I explained why I thought this idea was not interesting for me (in my context).

loki666:
no I meant, how does it runs on a Cortex ? is it powerful enough to generate the PCM stream without lags ?

troosh:
I understood, but I can not answer this question. Developers of [AYX-32](http://forum.tslabs.info/viewtopic.php?t=687) argued that the Cortex-M4 is not fast enough for FM synthesis (dual [YM2203](http://speccy.info/TurboSound_FM)  a.k.a. OPN). I also tried to understand how I can speed up the code and simplify it to rewrite it to ASM (possibly with using self-modifying code, ie JIT).

binarymaster:
@troosh ping, any news? Does this ready to merge, or you plan to do some more?

troosh:
Probably it is ready to merge. It's a pity that no one shared their experience with these patches.
I plan to continue optimizing, but I do not know when I can start this, agian.

Malvineous:
Does this belong in AdPlug, or should it really be merged into the upstream NukedOPL synth, and then copied here?  Not sure that having the AdPlug version of the NukedOPL synth diverge from the "official" one is a good idea.  It seems that these changes are substantial enough that it would be better for them to go into the official codebase and then we just keep in sync with that.

Malvineous:
I'm going to close this as WONTFIX, not because I don't appreciate the work that went into it, but because I don't want AdPlug to deviate from the upstream code.  I recommend getting these changes applied to the Nuked synth itself, which means we can then copy the whole updated version of the synth into AdPlug.

PR #63, opened 2018-01-23
Minor fixes

SoapGentoo:
* Fix for minor arches on Gentoo
* Fix for building out-of-source


SoapGentoo:
@Malvineous thanks for merging, would it be possible that you cut a 2.3.1 release so we can update Gentoo without running autoreconf to regenerate the configure script? Thanks!

Malvineous:
I have no problem with an updated release, but I'm a little short on time at the moment.  Any takers?

SoapGentoo:
@Malvineous I would gladly help, but I'm only a contributor

Malvineous:
Hopefully that problem has been fixed then :)

PR #64, opened 2018-04-20
`$<` is not guaranteed to work in ordinary make rules

SoapGentoo:
https://www.gnu.org/software/autoconf/manual/autoconf.html#g_t_0024_003c-in-Ordinary-Make-Rules

`$<` does not work in ordinary rules portably, and FreeBSD's `pmake` chokes on it.


SoapGentoo:
@Malvineous comment added

PR #69, opened 2018-10-10
Build NuGet packages with VS2017 support

### LINKED ISSUE #68: Support VS2017 using v141 toolset
Currently appveyor only creates NuGet packages containing the v140 toolset (VS2015). It would be nice to support VS2017 with the v141 platform toolset.

There is a dependency on libbinio which also supports up to v140. https://github.com/adplug/libbinio/pull/7 would include v141 packages in libbinio.

Herschel:
Adds VS platform toolset v141 to the AppVeyor builds which allows building on VS2017 with the v141 toolset. Fixes #68

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,5 @@
 version: 2.3.2-beta{build}
+image: Visual Studio 2017
 
 install:
 - ps: >-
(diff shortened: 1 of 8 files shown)


Malvineous:
Many thanks!

Issue #70, opened 2018-11-01
User configurable FREQ_OFFSET? 


PalMusicFan:
The surroundopl is really a fantastic enhancement of the OPL2 mono music formats. However, as I found in "surroundopl.h", the FREQ_OFFSET of the right channel is not configurable in run-time. Though "128.0" performs well in most of the tracks, we have found some percussion sound effects not compatible with it. Please refer to the DRO clip attached. As we heard, there is a percussion sound effect sounds well in the left channel while there is a distortion in the right channel. 

Although "384.0" solves this problem and the stereo effect is still nice, I have no idea whether this value is universally suitable. So, is it better to make it user configurable?

Thanks so much!  

[pal_percussion.zip](https://github.com/adplug/adplug/files/2536861/pal_percussion.zip)

PalMusicFan:
Actually we found this percussion sound effect requires an more accurate OPL frequency.

Malvineous:
I'm not sure that it's worth the effort of making it user configurable, however I'm open to accepting patches.  I will try to get some time to listen to the DRO clip as it sounds like the problem is that the frequency transposition isn't working correctly, which means changing the offset will just move the problem note to some other frequency instead.  So better to fix the frequency calculations so the problem goes away entirely.

In my own early experiments, values much below 110 made the effect difficult to notice, and values above 150 started to introduce dissonance making the songs sound off-key, so the actual range of values that sounded good was so small I didn't think it would be useful for end-users to be able to change it.

PalMusicFan:
Thanks for your reply! 
The following data is the result of my test I just made: 
182 and below: distortion 
183 to 364: not correct, however similar as a "spring sound"
365 and above: correct 
Actually I was surprised that changing FREQ_OFFSET value did not make a "linear change" of the sound.  When changing the OPL core frequency, the sound changes linearly and periodically, while only 49716Hz making the sound totally normal.

PalMusicFan:
In my test, values much below 100 made the sound right channel sound off-key (an elevated tone),  and an extremely high value (like "1024") made the surround effect difficult to notice, but "384" was acceptable.

Malvineous:
[...]

PR #73, opened 2019-01-18
Added user configurable frequency offset for surroundopl. 

PalMusicFan:
Hello! 

Please refer to the issue of "User configurable FREQ_OFFSET? #70 ". For some music tracks, perhaps it is necessary to set a surroundopl frequency offset value greater than 128 in order to avoid "distortion" of certain voice. So I think a user configurable frequency offset for surroundopl is pretty useful. 

I have added the feature of user configurable freq offset. Now the default frequency offset value is still defined by FREQ_OFFSET, which is 128. Frontend developers could set a new frequency offset value with something like: 
`opl->set_offset(cfg.offset);`

### DIFF
--- a/src/opl.h
+++ b/src/opl.h
@@ -60,6 +60,9 @@ class Copl
 
   // Emulation only: fill buffer
   virtual void update(short *buf, int samples) {}
+  
+  // Set surroundopl offset
+  virtual void set_offset(double offset) {}
 
  protected:
   int		currChip;		// currently selected OPL chip number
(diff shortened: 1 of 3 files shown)


PalMusicFan:
This change is compatible with current frontends (at least current adplay-unix).  
This is a demo of the new frontend of adplay-unix:
https://github.com/PalMusicFan/adplay-unix/tree/user-configurable-freq-offset
I have added a new parameter of "--offset" (or "-w", sorry I have no idea which character should be used for this, so I just simply choosed the next unoccupied one) to the player and the default offset value is also 128.

binarymaster:
Many thanks!

PR #74, opened 2019-02-16
CNemuopl::init should call OPL3_Reset.

adambiser:
Otherwise the emulator doesn't reset when the song rewinds, etc.

Similar code was already in CEmuopl::init.


binarymaster:
LGTM. Thanks! :+1:

Issue #66, opened 2018-07-30
Support updated RAD format


ghost:
[Reality Adlib Tracker 2](http://www.3eality.com/productions/reality-adlib-tracker) was released a couple months ago with an updated file format.

C++ source for a player and example tracks are included and [it seems fairly well documented](http://www.3eality.com/productions/reality-adlib-tracker/docs) (ctrl+f "coding notes").

devinacker:
I've actually made a branch of AdPlug a few months ago that added the RAD2 replayer, which I also modified to also support the original RAD format (instead of having them get converted to the generic Protracker-like format like they are currently).

I didn't release it at first because the replay code that came with the tracker didn't seem to actually specify any kind of license, and I don't remember if I ever contacted Reality about it (or if they ever added a license in the meanwhile...) But if that gets cleared up I can go ahead and open a PR, since it's basically already finished.

AliceLR:
I contacted Reality a few days ago about their license and SHAYDE replied: "All source code in the RAD package is public domain."

devinacker:
Excellent, I can try to get a PR ready soon then.

PR #75, opened 2019-02-16
replace original RAD loader with new RAD2 player

devinacker:
Per #66, this replaces the original RAD loader with a new loader/player which is basically just a wrapper around the replayer source included with the actual tracker, which I also modified to support the old v1 format (the source included with the tracker only supports v2's format even though I'm pretty sure the tracker itself can load both...)

Tested with the RAD v2 demo tunes plus pretty much every old RAD file I had on hand (i.e. the ones on ModLand). Didn't notice any old problems, although I guess `test/ALLOYRUN.ref` would need to be regenerated to match the new player as well.


Malvineous:
Definitely update the .ref file as part of the PR so we can be sure that the rest of the tests work too!

devinacker:
Done, hoping the CI builds are happy with it...

AliceLR:
There's a bug I reported to Reality in the original RAD v2 player that might be worth looking into here too. [This RAD v1 track](https://www.3eality.com/Public/Modules/RAD-v1/Void/mindflux.rad) has a `D31` effect at the end of order 12 that should make it jump to the middle of order 13, but this player (when modified to play v1 tracks) instead plays the entirety of order 13 very quickly. This fix worked for us: add this immediately after `Track = GetTrack();` at the end of `RADPlayer::PlayLine`:

```
        if(Line > 0)
            Track = SkipToLine(Track, Line, false);
```

You also might have to modify `RADPlayer::SkipToLine` to handle v1 tracks.

devinacker:
Done, also fixed some `#include`s to hopefully unbreak non-Visual Studio builds (I really hate how lenient it is with those...)

binarymaster:
Since this PR adds support for RAD v2 format, it would be nice to add this kind of file to testcases.

`mindflux.rad` also can be added to tests since it requires special handling in the code.

AliceLR:
I found this in the RAD v2 manual yesterday and I think it's also worth bringing up:

> One notable omission from earlier versions of RAD was the arpeggio effect (0xx).
Now you can use riffs to simulate  this effect.

I haven't looked for a test case for this yet.

AliceLR:
(I removed my previous comment re: arpeggios. I misunderstood the RAD v2 manual; the v1 format appears to have never had them.)

AliceLR:
I tried the rest of the .RAD files on Reality's public file system and found some more tracks that might be of interest:

[This track](https://www.3eality.com/Public/Modules/RAD-v1/_unknown/blue_ad.rad) contains invalid notes in numerous repeating junk patterns that aren't referenced by the order list. The v2 tracker interprets them as note off. This track otherwise loads and plays fine.

[These](https://www.3eality.com/Public/Modules/RAD-v1/_unknown/rip3.rad) [four](https://www.3eality.com/Public/Modules/RAD-v1/_unknown/rip4.rad) [tracks](https://www.3eality.com/Public/Modules/RAD-v1/_unknown/rip5.rad) [each](https://www.3eality.com/Public/Modules/RAD-v1/_unknown/rip8.rad) have a byte truncated from the final note of their final pattern. The tracker will load them and play them with mostly no issue, though the instrument number and effect of the final note will be junk.

devinacker:
I'm traveling for the weekend but I'll check those out in a few days.

devinacker:
Those should play back okay now.

(The "fix" for a truncated last note is a little cheesy - it just forces an extra null byte at the end of the file to form a valid final note, but shouldn't affect patterns that have any more missing data than that)

AliceLR:
FYI there are some more bugs another user of the RAD v2 player found:

* The validate function checks bit 6 as the BPM flag instead of bit 5.
* The player skips 7 bytes total for MIDI instruments instead of 6 (should be `s += 5` since the algorithm byte was already skipped).
* The player will not execute a command on the same row+channel a MIDI instrument is played.

My fixes are in the [PR referenced above](https://github.com/AliceLR/megazeux/pull/188) but I haven't fully tested them yet.

devinacker:
Thanks for the info, I'll check that stuff out probably sometime next week since I'm about to be traveling again.

Malvineous:
How's this going?  Is it ready to merge or is there still more work to do?

devinacker:
I'm alright with it being merged as-is, assuming no other issues have been found with the original replayer.

AliceLR:
FWIW I haven't found any issues with the original player since my last comment.

Malvineous:
That's great, thanks!  I've merged it into master now.

PR #77, opened 2019-02-21
Moved XSM OPL initialization code from load into rewind. 

adambiser:
Same reason as pull request #76.


adambiser:
The original code does not call opl->init().

I've looked many of the other CPlayer classes and they do call opl->init() in their rewind method.  However, adding this causes the eXtra Simple Music test to fail.

Should opl->write(1, 32) also be added also?  I saw that most CPlayer classes have this line also.

Malvineous:
Writing `32` to register `1` switches the chip into OPL3 mode, enabling some extra waveforms and other features.  Without this, the chip runs in OPL2 mode.  Whether you include this line or not depends on whether the file format is designed to run on an OPL2 or an OPL3.

If you have fixed the player by calling `opl->init()` then it's possible that the data now being produced is more correct than the test data, so it's no problem to include updated test data in the commit.

adambiser:
Isn't reg 32 the waveform select, not OPL3 mode which is setting reg 105 to 1?

Where can I find the test data?

Malvineous:
Sorry you're right, it is the Waveform Select Enable bit in register 1, not the OPL3 bit.  But it's still generally up to the file format whether the waveforms are used or not, although nearly everything uses them.  In theory it shouldn't matter if you enable it (so long as the actual waveform bits are zeroed out as they are by `init()`) because any format that doesn't support the Waveform Select Enable bit presumably isn't writing anything to the Waveform Select registers either, so the default zero values there should produce the same sound.

It's been a while but there is a folder for tests and a failed test should produce a file that you can rename to make it the proper test data.  I think I added some info to the README about how it works last time I had to update the tests.

adambiser:
OK, I added opl->write(1, 32) and updated the test ref file as well.

PR #78, opened 2019-02-22
Attempt at fixing ADL playback issues

adambiser:
This is an attempt to fix some problems mentioned in issue #6 .

The driver channels are now cleared in rewind (callback 8).  Before this change, values for previous subsongs could be held over into the new subsong when the previous subsong had a channel active that the new
subsong did not leaving a non-null dataptr value on an inactive channel which would then be processed in executePrograms.

Both rewind and update perform a check to be sure no subsong < 2 can be selected.  This is to solve the access violation that would occur when rewind() was called before update() and cursubsong would be -1.

Removed the update call from the end of rewind.  This would cause update to occur twice in the first update call (unless rewind was called first).  **This affects the test results because there is now another r72.00 near the beginning.**


adambiser:
Still a work in progress.

adambiser:
OK, I think the latest commit does what is expected.

The tests still fail because of the reason mentioned above.  Will update the tests if the code looks good.

adambiser:
I have posted a project that uses the AdPlug library with the changes made for this pull request [here](https://github.com/adambiser/agk-adlib-plugin/tree/master/Examples/PlayAdlib) if you want to see them "in action".

src/adl.cpp
```diff
@@ -2435,7 +2436,8 @@ bool CadlPlayer::load(const std::string &filename, const CFileProvider &fp)
   file_data = new uint8 [file_size];
   f->readString((char *)file_data, file_size);
 
-  _driver->callback(8, int(-1));
+  // Moved into rewind.
+  //_driver->callback(8, int(-1));
```
binarymaster: You can just remove these comments, they're obvious and can be easily seen in the patch diff.

adambiser: Done.

binarymaster:
Thanks!

PR #82, opened 2019-05-05
Unaligned access

SoapGentoo:
Hi @Malvineous 
in order to make debugging on PPC easier, I'm trying to clean up all the UBSAN warnings. It seems that the codebase does a lot of unaligned access which is undefined behaviour in C (and theoretically not even possible in C++, as you would have to break the strict aliasing rules before you even get to type pun). The recommended approach instead is to use `memcpy`, which every modern compiler can optimize to the best type of load on the target platform (on x86, GCC and Clang optimize this to a standard `mov` usually).

See also:

1. https://blog.regehr.org/archives/1520
2. https://pzemtsov.github.io/2016/11/06/bug-story-alignment-on-x86.html
3. https://blog.quarkslab.com/unaligned-accesses-in-cc-what-why-and-solutions-to-do-it-properly.html

### DIFF
--- a/src/herad.cpp
+++ b/src/herad.cpp
@@ -37,6 +37,7 @@
 #include <stdio.h>
 
 #include "herad.h"
+#include "load_helper.h"
 
 #ifdef DEBUG
 #include "debug.h"
@@ -86,7 +87,10 @@ bool isHSQ(uint8_t * data, int size)
 		#endif
 		return false;
 	}
-	if ( *(uint16_t *)(data + 3) != size )
+
+	const uint16_t temp_size = u16_unaligned(data + 3);
+
+	if ( temp_size != size )
 	{
 		#ifdef DEBUG
 		AdPlug_LogWrite("HERAD: Is not HSQ, wrong compressed size.\n");
@@ -148,7 +152,7 @@ uint16_t HSQ_decompress(uint8_t * data, int size, uint8_t * out)
 		// get next bit of the queue
 		if (queue == 1)
 		{
-			queue = *(uint16_t *)src | 0x10000;
+			queue = u16_unaligned(src) | 0x10000;
 			src += 2;
 		}
 		bit = queue & 1;
@@ -164,7 +168,7 @@ uint16_t HSQ_decompress(uint8_t * data, int size, uint8_t * out)
 			// get next bit of the queue
 			if (queue == 1)
 			{
-				queue = *(uint16_t *)src | 0x10000;
+				queue = u16_unaligned(src) | 0x10000;
 				src += 2;
 			}
 			bit = queue & 1;
@@ -174,7 +178,7 @@ uint16_t HSQ_decompress(uint8_t * data, int size, uint8_t * out)
 			{
 				// count = next 3 bits of the input
 				// offset = next 13 bits of the input minus 8192
-				count = *(uint16_t *)src;
+				count = u16_unaligned(src);
 				offset = (count >> 3) - 8192;
 				count &= 7;
 				src += 2;
@@ -194,7 +198,7 @@ uint16_t HSQ_decompress(uint8_t * data, int size, uint8_t * out)
 				// count = next bit of the queue * 2 + next bit of the queue
 				if (queue == 1)
 				{
-					queue = *(uint16_t *)src | 0x10000;
+					queue = u16_unaligned(src) | 0x10000;
 					src += 2;
 				}
 				bit = queue & 1;
@@ -202,7 +206,7 @@ uint16_t HSQ_decompress(uint8_t * data, int size, uint8_t * out)
 				count = bit << 1;
 				if (queue == 1)
 				{
-					queue = *(uint16_t *)src | 0x10000;
+					queue = u16_unaligned(src) | 0x10000;
 					src += 2;
 				}
 				bit = queue & 1;
@@ -233,7 +237,7 @@ uint16_t SQX_decompress(uint8_t * data, int size, uint8_t * out)
 	uint8_t * dst = out;
 	bool done = false;
 
-	*(uint16_t *)dst = *(uint16_t *)src;
+	std::memcpy(dst, src, sizeof(uint16_t));
 	src += 6;
 	uint16_t queue = 1;
 	uint8_t bit, bit_p;
@@ -243,7 +247,7 @@ uint16_t SQX_decompress(uint8_t * data, int size, uint8_t * out)
 		queue >>= 1;
 		if (queue == 0)
 		{
-			queue = *(uint16_t *)src;
+			queue = u16_unaligned(src);
 			src += 2;
 			bit_p = bit;
 			bit = queue & 1;
@@ -264,7 +268,7 @@ uint16_t SQX_decompress(uint8_t * data, int size, uint8_t * out)
 				queue >>= 1;
 				if (queue == 0)
 				{
-					queue = *(uint16_t *)src;
+					queue = u16_unaligned(src);
 					src += 2;
 					bit_p = bit;
 					bit = queue & 1;
@@ -282,7 +286,7 @@ uint16_t SQX_decompress(uint8_t * data, int size, uint8_t * out)
 					queue >>= 1;
 					if (queue == 0)
 					{
-						queue = *(uint16_t *)src;
+						queue = u16_unaligned(src);
 						src += 2;
 						bit_p = bit;
 						bit = queue & 1;
@@ -303,7 +307,7 @@ uint16_t SQX_decompress(uint8_t * data, int size, uint8_t * out)
 				}
 				break;
 			case 2:
-				count = *(uint16_t *)src;
+				count = u16_unaligned(src);
 				offset = (count >> data[5]) - (1 << (16 - data[5]));
 				count &= (1 << data[5]) - 1;
 				src += 2;
@@ -335,7 +339,7 @@ uint16_t SQX_decompress(uint8_t * data, int size, uint8_t * out)
 			queue >>= 1;
 			if (queue == 0)
 			{
-				queue = *(uint16_t *)src;
+				queue = u16_unaligned(src);
 				src += 2;
 				bit_p = bit;
 				bit = queue & 1;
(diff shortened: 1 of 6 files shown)


src/mid.cpp
```diff
@@ -300,7 +300,8 @@ bool CmidPlayer::load(const std::string &filename, const CFileProvider &fp)
     uint32_t size;
 
     f->readString((char *)s, 6);
-    size = *(uint32_t *)s; // size of FILE_OLDLUCAS
+    // need to perform unaligned access
+    memcpy(&size, s, sizeof(uint32_t)); // size of FILE_OLDLUCAS
```
Malvineous: This will still yield different results on big endian vs little.  Might be good to make it endian neutral at the same time.

src/herad.cpp
```diff
@@ -148,7 +153,10 @@ uint16_t HSQ_decompress(uint8_t * data, int size, uint8_t * out)
 		// get next bit of the queue
 		if (queue == 1)
 		{
-			queue = *(uint16_t *)src | 0x10000;
+			// need to perform unaligned access
+			uint16_t temp;
+			memcpy(&temp, src, sizeof(uint16_t));
+			queue = temp | 0x10000;
```
Malvineous: While I'm not opposed to these changes in principle, they do make the code a little more verbose.  Is there any way you could refactor things so the `memcpy` is not necessary, e.g. by making `src` a `uint16_t` array from the start, so it can be accessed directly as `*src`, and `src++` instead of `src += 2`?

Failing that, I wonder whether a macro or inline function could be used to hide the temporary variable and make the code a bit clearer?

SoapGentoo: The problem is: Do you have any guarantees that the incoming stream will always be aligned? Say for instance we have the following datastream:

```
|-------- 4 bytes --------|- 1 bytes -|---- 2 bytes ----|
```
then the last two bytes will not be aligned correctly anymore, regardless of the fact that the initial array might have been 4-byte aligned. My suggestion is to create littler getters/setters like
```
uint16_t get_uint16_t(const unsigned char* src, endianness = LITTLE)
{
  uint16_t res;
  memcpy(&res, src, sizeof(res));
  return res;
}

void set_uint16_t(uint16_t& dest, const unsigned char* src, endianness = LITTLE)
{
  dest = get_uint16_t(src, endianness);
}
```
which allows cleaning up the whole mess in one go, and memcpy is nicely tucked away in one function. What do you think?

src/mid.cpp
SoapGentoo: Correct, I wasn't going to fix the endianness issues with this PR yet

src/herad.cpp
Malvineous: You are correct that if the incoming stream has different bit-widths then you will need something like this.

However all your proposed changes in `herad.cpp` are `uint16_t` which makes me wonder whether you could just change `src` to be a `uint16_t` array, and load it element by element with `readInt(2)`.  This way the platform may allocate it as an array of 32-bit fields in order to allow direct element access, but this won't matter as you're never accessing the memory directly if you treat it as a normal array.  It will also fix endian issues as the bytes will be loaded in the correct order from the start.  For that matter you could forget the size-specific component and simply allocate it as an array of `unsigned int`.  However I haven't looked at the code in detail so not sure whether this will work.

This does beg the question that if C/C++ cannot do unaligned memory accesses, what actually happens if you have an array of `uint16_t`?  Is it padded in memory or does the compiler generate extra instructions to allow unaligned accesses?

src/mid.cpp
Malvineous: That's fine, but I suspect the endian issues are going to be significant enough that it may involve a second rewrite, so might be a good idea to try to fit both in at the same time if possible, just to save some effort.

src/herad.cpp
SoapGentoo: @Malvineous checkout line 160/168:
```
// copy next byte of the input to the output
*dst++ = *src++;
```
so clearly `src` points to a stream of intermingled `uint8_t`s and `uint16_t`s, which means all alignment guarantees will be out the window at *some point* during this parse loop.

**Valid** C/C++ code will never do unaligned access, because the compiler will insert padding wherever necessary (which is why you should avoid packed structs like the plague). For instance, a normal `uint16_t arr[N]` will **always** be 2 byte aligned on literally every platform but the msot crude 8-bit CPUs. The problem here is that you have a stream of lots of different things, and at some point you're treating a bunch of 4 bytes like an `uint32_t` and doing pointer casts, which violates 1) strict aliasing in C/C++ 2) more often than not violates the alignment requirements on C. The extreme example here is SPARC: I tried, this code crashes on SPARC due to a `SIGBUS`, as SPARC *does not support unaligned access*.

Malvineous: Yes I can see my thought of changing `src` to be a `uint16_t` array is probably more trouble than it's worth, if the compiler optimises away the `memcpy` call.

Your proposed `get_uint16_t` function looks good, although declaring it `inline` might help optimisations.  Perhaps we could call the function something like `u16_unaligned()` just as a reminder why it's there?  I'm the sort of programmer who is likely to replace the function with typecasts one day thinking they are more efficient, and completely forgetting about the alignment issues.  This will also avoid the need to comment each use.

What do you think?

SoapGentoo: Sure, I while work on that then. I can only work on it on the weekend, and will change it as discussed.
(sidenote: `inline` has negligible effect on whether the function actually gets inlined. `inline` is important in C++ for reasons of ODR (multiple, but equal, definitions of the same function), whereas in C the semantics are different but still do not really affect inlining but whether an eternal symbol is emitted at all).

SoapGentoo:
@Malvineous please have a look again, modifications as discussed. I've bumped the build system to C++11 unconditionally, in order to get the benefit of `cstdint` and `enum class`.

SoapGentoo:
@Malvineous ping

src/herad.cpp
```diff
@@ -86,7 +87,10 @@ bool isHSQ(uint8_t * data, int size)
 		#endif
 		return false;
 	}
-	if ( *(uint16_t *)(data + 3) != size )
```
SoapGentoo: @Malvineous this already invokes C++11, whether you specify it or not. C++98 does not officialy recognize `int*_t` types, so by using these, you are already implicitly relying in C++11

SoapGentoo: With `CXXFLAGS="-std=c++98 -Wpedantic"`

    /usr/include/libbinio/binio.h:99:16: warning: ISO C++ 1998 does not support ‘long long’ [-Wlong-long]
       typedef long long Int;

SoapGentoo:
@Malvineous please have a look again. I have renamed the unaligned loaders and removed the explicit dependency on C++11. That said, I still think depending on C++11 is a non-issue because even obscure platforms have the basic C++11 features I am requesting here, we are relying on it anyways (`long long` and `uint*_t`). Can you give me an example of a common platform where adplug is being used without a C++11 compiler?

SoapGentoo:
@Malvineous ping

SoapGentoo:
@Malvineous ping

src/herad.cpp
```diff
@@ -86,7 +87,10 @@ bool isHSQ(uint8_t * data, int size)
...
+
+	const uint16_t temp_size = u16_unaligned(data + 3);
+
+	if ( temp_size != size )
```
binarymaster:
```suggestion
	if (temp_size != size)
```
I added surrounding spaces inside brackets before only for better readability, with these changes they are not needed anymore. :slightly_smiling_face:

SoapGentoo: @binarymaster the really correct way of approaching this is to use something like `clang-format` to enforce a consistent formatting :wink:.

SoapGentoo:
@Malvineous ping

src/herad.cpp
Malvineous: These types are definitely supported by earlier language versions, you just have to `#include <stdint.h>` for C or `#include <cstdint>` for C++ to use them.

Malvineous:
I can't say for sure which platforms don't have C++11 compilers because all the platforms I use do, but people have complained in the past (before my time) when we introduced new features that weren't available on their platforms.  I recently removed support for 8-bit audio for example, and it was only a few weeks later someone wanted it added back in again!

There are unofficial ports for other platforms like the Amiga, and I have no idea whether they have C++11 compilers or not.  If we make the switch to require C++11, it has to be for a *very* good reason, not just a bit of convenience.

src/load_helper.h
```diff
@@ -0,0 +1,58 @@
+#ifndef ADPLUG_LOAD_HELPER_H
+#define ADPLUG_LOAD_HELPER_H
+
+#include <cstring>
+#include <stdint.h>
```
Malvineous: Should this be `<cstdint>` seeing as we're in C++ here?

Malvineous:
Looks good, I'm happy with this now.  If you can make those last couple of changes I'll merge it - thanks!

src/load_helper.h
SoapGentoo: no, and that's because `<cstdint>` is a C++11 header. The fact that `<stdint.h>` works is purely coincidence, because C++98 never mentions `<cstdint>` or `<stdint.h>`, and that is due to the presence of C99 headers. Only C++11 uses C99 as a referential standard. The fact that this works is coincidence and not design (and fails on MSVC for instance).

src/herad.cpp
SoapGentoo: Again, C++98 has no notion of `<cstdint>`, and the fact that it works is more an accident than something you can rely on. Many old compilers will break when including `<cstdint>`, they just happen to work with `<stdint.h>` due to the presence of C99 headers on the system.

Furthermore, `long long` is not defined in C++98. C99 introduced `long long` and C++11 references C99 as a base C standard. What I'm trying to get at here is that you're using C++11 (and C99) features implicitly already.

src/load_helper.h
SoapGentoo: This won't work on old Apple compilers, that have C99 headers (`<stdint.h>`) but not C++11 headers (`<cstdint>`) yet.

Malvineous: Oh you're right, my mistake.  I have only used `stdint.h` myself and I thought I must have been doing it wrong but that would be why I never saw `cstdint` before!  No argument about using C99 features already, but the fact of the matter is that the code compiles for everyone now, whereas if we make C++11 mandatory I am absolutely sure it will break for *someone*.

PR #83, opened 2019-05-07
Allow 8-bit playback

omicronapps:
Currently reducing from 16-bit to 8-bit samples does not work, since 'mixbuf1' is used for both the left channel source buffer (with TYPE_DUAL_OPL2) and the stereo source destination buffer. Specifically, 'mixbuf1' array is already used by 'tempbuf2' pointer, but if 'use16bit' is false then 'mixbuf1' is also used by 'outbuf' pointer. This causes data to be overwritten when copying data from 'tempbuf2' to 'outbuf', as both are pointing to 'mixbuf1'.

To resolve this, add a new array 'mixbuf2' for 8-bit output to 'outbuf' pointer.

### DIFF
--- a/src/emuopl.cpp
+++ b/src/emuopl.cpp
@@ -39,6 +39,7 @@ CEmuopl::~CEmuopl()
   if(mixbufSamples) {
     delete [] mixbuf0;
     delete [] mixbuf1;
+    delete [] mixbuf2;
   }
 }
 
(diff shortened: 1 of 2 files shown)


omicronapps:
@Malvineous @binarymaster @SoapGentoo 
Please review changes to enable 8-bit audio rendering.

Malvineous:
Thanks! Merged in ee55159cfff41eceb887f4bc303ab32f0c3b158f

Issue #91, opened 2019-08-09
Double free in Cu6mPlayer::~Cu6mPlayer()


fcambus:
Hi,

While fuzzing AdPlug with American Fuzzy Lop, I found a double free in in Cu6mPlayer::~Cu6mPlayer(), in src/u6m.h L42.

Attaching a reproducer (gzipped so GitHub accepts it): [double-free.gz](https://github.com/adplug/adplug/files/3485266/double-free.gz)

```
=================================================================
==16659==ERROR: AddressSanitizer: attempting double-free on 0x629000000200 in thread T0:
    #0 0x4f75a2 in operator delete[](void*) (/home/fcambus/adplay-asan/adplay+0x4f75a2)
    #1 0x7f96ab794bb7 in Cu6mPlayer::~Cu6mPlayer() /home/fcambus/adplug/./src/u6m.h:42:21
    #2 0x7f96ab794bf8 in Cu6mPlayer::~Cu6mPlayer() /home/fcambus/adplug/./src/u6m.h:41:5
    #3 0x7f96ab63b66a in CAdPlug::factory(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, Copl*, CPlayers const&, CFileProvider const&) /home/fcambus/adplug/src/adplug.cpp:187:2
    #4 0x4fcd62 in play(char const*, Player*, int) /home/fcambus/adplay-unix/src/adplay.cc:309:11
    #5 0x4fbaf4 in main /home/fcambus/adplay-unix/src/adplay.cc:544:5
    #6 0x7f96aae7909a in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2409a)
    #7 0x41f759 in _start (/home/fcambus/adplay-asan/adplay+0x41f759)

0x629000000200 is located 0 bytes inside of 16735-byte region [0x629000000200,0x62900000435f)
freed by thread T0 here:
    #0 0x4f75a2 in operator delete[](void*) (/home/fcambus/adplay-asan/adplay+0x4f75a2)
    #1 0x7f96ab78d0a9 in Cu6mPlayer::load(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, CFileProvider const&) /home/fcambus/adplug/src/u6m.cpp:87:7
[...]

PR #97, opened 2019-08-21
Fix issue #91

debrouxl:

PR #103, opened 2019-08-22
Add test case to trigger double free in Cu6mPlayer::~Cu6mPlayer() (Is…

fcambus:
…sue #91)


Malvineous:
I tried merging the fix in #97 however this test still fails afterwards.  I am not sure whether PR #97 isn't fixing the problem, or whether this PR isn't correctly testing the problem.  Are you able to take a look?

Malvineous:
Closing as fix came in with 8abb9328bf27dcbdafc67ade3e75af0ffd8f7633 with a test case.

PR #106, opened 2019-12-19
woodyopl: Moved generator_add

KyleJ61782:
Moved generator_add from the woodyopl header to the source file as it's
not used outside of the source file and having it in the header results
in compiler warnings.

### DIFF
--- a/src/woodyopl.cpp
+++ b/src/woodyopl.cpp
@@ -33,6 +33,7 @@
 #include <string.h> // memset
 #include "woodyopl.h"
 
+static Bit32u generator_add;	// should be a chip parameter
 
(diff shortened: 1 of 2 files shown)


Malvineous:
Thanks for the contribution but I don't think it would be right to merge this change here.  The WoodyOPL implementation is more or less a straight copy from the DOSBox source code, with only the bear minimum required changes to get it to work with AdPlug.

If we start merging changes like this in, they will be lost again the next time we copy across the latest version from the DOSBox code.

I think this patch is better submitted to DOSBox instead, and once incorporated there we can just do another update here to get it as well.  That way it's easier to remain in sync with the true owner of the code.

Issue #102, opened 2019-08-21
Wild write (memory corruption) at user-controlled address in CdfmLoader::load()


debrouxl:
Many offending samples: [wild_write.tar.gz](https://github.com/adplug/adplug/files/3527212/wild_write.tar.gz).

The wide variety of addresses for the faulting address caught my eye early on. I checked and played with the contents of 7 files: **all of them contain at least an occurrence of the sequence of bytes corresponding to the write fault address** (as little-endian on this x86_64 computer, as expected).

**Hex-editing the bytes of the appropriate occurrence changes the fault address** and type in a way I haven't fully understood; I can see that the write address range achievable through some samples is 47 LSBs on this x86_64 computer, while some samples can only control a range of e.g. 31 LSBs. Edit a byte too far and the sample decays to a wild OOB read whose address is also partially controlled and which also usually crashes. With samples which control 47 LSBs, the address can be pointed anywhere.

The value written to the user-specified fault address is also partially user-specified (see below), so this looks to be between a write-where primitive and a full-blown write-what-where primitive.

Output given for two samples, the dozens of others samples produce similar output:
```
(47 LSBs address range control capability)
AddressSanitizer:DEADLYSIGNAL
=================================================================
==9991==ERROR: AddressSanitizer: SEGV on unknown address 0x297213b53626 (pc 0x7f06d0e47a41 bp 0x7ffca23f1130 sp 0x7ffca23f0fc0 T0)
==9991==The signal is caused by a WRITE memory access.
    #0 0x7f06d0e47a40 in CdfmLoader::load(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, CFileProvider const&) .../adplug/src/dfm.cpp
    #1 0x7f06d0dfdc3a in CAdPlug::factory(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, Copl*, CPlayers const&, CFileProvider const&) .../adplug/src/adplug.cpp:182:13
    #2 0x4fd913 in play(char const*, Player*, int) .../adplay-unix/src/adplay.cc:309:11
    #3 0x4fd913 in main .../adplay-unix/src/adplay.cc:554
    #4 0x7f06d020c09a in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2409a)
    #5 0x4208b9 in _start (.../adplug_prefix/bin/adplay+0x4208b9)

AddressSanitizer can not provide additional info.
[...]

Issue #85, opened 2019-08-06
Heap-based buffer overflow in CxadbmfPlayer::__bmf_convert_stream()


fcambus:
Hi,

While fuzzing AdPlug with American Fuzzy Lop, I found a heap-based buffer overflow in CxadbmfPlayer::__bmf_convert_stream(), in src/bmf.cpp L430.

Attaching a reproducer (gzipped so GitHub accepts it): [test01.bmf.gz](https://github.com/adplug/adplug/files/3471523/test01.bmf.gz)

```
=================================================================
==6298==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x63000000f8dd at pc 0x0000004c6aaf bp 0x7ffd579ffc90 sp 0x7ffd579ff440
WRITE of size 6 at 0x63000000f8dd thread T0
    #0 0x4c6aae in __asan_memset (/home/fcambus/reps/adplay+0x4c6aae)
    #1 0x7face9bf1bdd in CxadbmfPlayer::__bmf_convert_stream(unsigned char*, int) /home/fcambus/adplug/src/bmf.cpp:430:5
    #2 0x7face9bf1a0a in CxadbmfPlayer::xadplayer_load() /home/fcambus/adplug/src/bmf.cpp:189:12
    #3 0x7face9d5d88a in CxadPlayer::load(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, CFileProvider const&) /home/fcambus/adplug/src/xad.cpp:93:8
    #4 0x7face9bdf1d5 in CAdPlug::factory(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, Copl*, CPlayers const&, CFileProvider const&) /home/fcambus/adplug/src/adplug.cpp:169:10
    #5 0x4fcd62 in play(char const*, Player*, int) /home/fcambus/adplay-unix/src/adplay.cc:309:11
    #6 0x4fbaf4 in main /home/fcambus/adplay-unix/src/adplay.cc:544:5
    #7 0x7face941d09a in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2409a)
    #8 0x41f759 in _start (/home/fcambus/reps/adplay+0x41f759)

0x63000000f8dd is located 6061 bytes to the right of 56624-byte region [0x630000000400,0x63000000e130)
allocated by thread T0 here:
[...]

PR #108, opened 2020-04-03
Add fuzzing samples from issues #85 to #102 as new tests

miller-alex:
Add a new test program for testing fuzzing samples. For now it simply tries to load and play test samples, but this can be extended to cover other parts of the code when samples exposing errors there surface. A a test case fails if the test sample leads to a crash or nonzero exit code (i.e., if an attempt to load a file fails by returning false, that's success).
All fuzzing samples provided by @fcambus and @debrouxl in the issues #85 to #102 are added as test cases to the new test program. Some of these seem to be redundant, especially @debrouxl often added a bunch of samples that trigger the same bug, but I added them all for reference. They can be cleaned up later if desired.


miller-alex:
Oh, I should mention that these tests may need a memory checker to detect all issues (compile with address sanitizer, run with valgrind, or similar). I didn't hook it up for CI or even add the tests to the windows build files.

Malvineous:
This is really good but unfortunately it really slows down the build process because the stress test takes so long to run.  Is there any way it can be sped up?  Maybe running some tests in parallel or something?

Malvineous:
I'm closing this issue because I have merged it into master, but if you are able to speed it up, please open a new PR with your improvements!  Many thanks :)

miller-alex:
Adding lots of test inevitably slows things down, and I've seen a new batch of samples has already been posted in #110, so it will get even worse. A few thoughts about speeding tests up:

* Parallelization helps a bit, but I don't intend to implement a parallel test harness with bells and whistles in stresstest. If you want to go that route, I'd suggest to remove the ability to run all samples from stresstest, making each sample an individual test case, and move that work to the test harness. No need to reinvent the wheel.
* The easiest way to speed tests up is to reduce the number of test samples. I've simply added all samples that were attached to the issues, bit I think a lot of them are redundant. (The most extreme example, #102 has 151 test samples, but I think it affected only 2 or 3 locations and was fixed with a single change.) Redundant samples may be removed or disabled by default.
* If the stresstest is too slow, it can be removed from the default test target and only be run on explicit request. After all, a memory checker is needed to make it really effective. Building everything with ASAN is not what you typically do, and running stresstest with valgrind is even slower. So maybe it's not that interesting to run it by default.

PR #109, opened 2020-04-03
Fix issues #85 to #102

miller-alex:
This fixes a truckload of invalid memory accesses and other crashes and undefined behavior in several players, including those reported in the issues #85 to #102.
This addresses only obviously invalid code in the files affected by those issues, didn't bother with other bugs or other files; there's probably still a lot of crashing bugs left to be fixed.


Malvineous:
Merged, many thanks :)

PR #111, opened 2020-05-12
Clean up bmf.cpp

miller-alex:
When I started working on the fixes that were merged recently, I did some additional formatting changes and refactoring for the first file I was working on. (I didn't do that for other files because there's a lot more important stuff to fix.) Since these commits introduce no functional change and don't fix anything I didn't include them in the other pull requests.

Some of the changes may be a matter of taste, but I think most of them improve readability of the code. Feel free to merge if you like the changes (or modify what you don't like or even close without merging). What do you think?

### DIFF
--- a/src/bmf.cpp
+++ b/src/bmf.cpp
@@ -95,13 +95,13 @@ bool CxadbmfPlayer::xadplayer_load()
 #endif
   if (tune_size < 6)
     return false;
-  if (!strncmp((char *)tune, "BMF1.2", 6))
+  if (!memcmp(&tune[0], "BMF1.2", 6))
   {
     bmf.version = BMF1_2;
     bmf.timer = 70.0f;
     ptr = 6;
   }
-  else if (!strncmp((char *)tune, "BMF1.1", 6))
+  else if (!memcmp(&tune[0], "BMF1.1", 6))
   {
     bmf.version = BMF1_1;
     bmf.timer = 68.5f;
@@ -119,34 +119,34 @@ bool CxadbmfPlayer::xadplayer_load()
     // It's unclear whether title/author can be longer than 35 chars. The
     // implementation reads an arbitrarily long NUL-terminated string and
     // truncates it to fit in a fixed 36 byte buffer. Doesn't make sense!
-    size_t len = strnlen((char *)tune + ptr, tune_size - ptr);
+    size_t len = strnlen((char *)&tune[ptr], tune_size - ptr);
     if (ptr + len == tune_size)
       return false;
-    if (len > 35)
+    if (len >= sizeof(bmf.title))
     {
-      memcpy(bmf.title, tune + ptr, 35);
-      bmf.title[35] = 0;
+      memcpy(bmf.title, &tune[ptr], sizeof(bmf.title) - 1);
+      bmf.title[sizeof(bmf.title) - 1] = 0;
     }
     else
-      memcpy(bmf.title, tune + ptr, len + 1);
+      memcpy(bmf.title, &tune[ptr], len + 1);
     ptr += len + 1;
 
-    len = strnlen((char *)tune + ptr, tune_size - ptr);
+    len = strnlen((char *)&tune[ptr], tune_size - ptr);
     if (ptr + len == tune_size)
       return false;
-    if (len > 35)
+    if (len >= sizeof(bmf.author))
     {
-      memcpy(bmf.author, tune + ptr, 35);
-      bmf.author[35] = 0;
+      memcpy(bmf.author, &tune[ptr], sizeof(bmf.author) - 1);
+      bmf.author[sizeof(bmf.author) - 1] = 0;
     }
     else
-      memcpy(bmf.author, tune + ptr, len + 1);
+      memcpy(bmf.author, &tune[ptr], len + 1);
     ptr += len + 1;
   }
   else
   {
-    strncpy(bmf.title, xad.title, 36);
-    strncpy(bmf.author, xad.author, 36);
+    strncpy(bmf.title, xad.title, sizeof(bmf.title));
+    strncpy(bmf.author, xad.author, sizeof(bmf.author));
   }
 
   // speed
@@ -169,21 +169,25 @@ bool CxadbmfPlayer::xadplayer_load()
     {
       if (iflags & (1 << (31-i)))
       {
-        if (tune_size - ptr < 24)
+        if (tune_size - ptr < sizeof(bmf.instruments[i]))
           return false;
-        strncpy(bmf.instruments[i].name, (char *)tune + ptr, 10);
-        // 11th char ignored since name must be 0-terminated
-        bmf.instruments[i].name[10] = 0;
-        memcpy(bmf.instruments[i].data, tune + ptr + 11, 13);
-        ptr += 24;
+        memcpy(bmf.instruments[i].name, &tune[ptr],
+                sizeof(bmf.instruments[i].name) - 1);
+        // last char ignored since name must be 0-terminated
+        bmf.instruments[i].name[sizeof(bmf.instruments[i].name) - 1] = 0;
+        memcpy(bmf.instruments[i].data,
+               &tune[ptr + sizeof(bmf.instruments[i].name)],
+               sizeof(bmf.instruments[i].data));
+        ptr += sizeof(bmf.instruments[i]);
       }
       else if (bmf.version == BMF1_1)
       {
-        memset(bmf.instruments[i].name, 0, 11);
-        memcpy(bmf.instruments[i].data, bmf_default_instrument, 13);
+        memset(bmf.instruments[i].name, 0, sizeof(bmf.instruments[i].name));
+        memcpy(bmf.instruments[i].data, bmf_default_instrument,
+               sizeof(bmf.instruments[i].data));
       }
       else
-        memset(bmf.instruments + i, 0, sizeof(bmf.instruments[i]));
+        memset(&bmf.instruments[i], 0, sizeof(bmf.instruments[i]));
     }
   }
   else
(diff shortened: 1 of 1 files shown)


src/bmf.cpp
```diff
@@ -83,23 +87,25 @@ bool CxadbmfPlayer::xadplayer_load()
...
+  if (xad.fmt != BMF)
     return false;
 
 #ifdef DEBUG
   AdPlug_LogWrite("\nbmf_load():\n\n");
 #endif
   if (tune_size < 6)
     return false;
-  if (!strncmp((char *)&tune[0],"BMF1.2",6))
+  if (!strncmp((char *)tune, "BMF1.2", 6))
```
Malvineous, line 98: This one can sometimes be preferable the way it was, as it makes it a bit clearer you're talking about offset 0 in the file.

src/bmf.cpp
```diff
@@ -83,23 +87,25 @@ bool CxadbmfPlayer::xadplayer_load()
...
   {
     bmf.version = BMF1_2;
     bmf.timer = 70.0f;
+    ptr = 6;
   }
-  else if (!strncmp((char *)&tune[0],"BMF1.1",6))
+  else if (!strncmp((char *)tune, "BMF1.1", 6))
   {
```
Malvineous, line 105: Same here.  It's probably a matter of taste, but as we're often looking at files in a hex editor when implementing this stuff I do like the added clarity you can get from `[0]`.

src/bmf.cpp
```diff
@@ -141,48 +145,46 @@ bool CxadbmfPlayer::xadplayer_load()
...
       if (iflags & (1 << (31-i)))
-	  {
+      {
         if (tune_size - ptr < 24)
           return false;
         strncpy(bmf.instruments[i].name, (char *)tune + ptr, 10);
         // 11th char ignored since name must be 0-terminated
         bmf.instruments[i].name[10] = 0;
-        memcpy(bmf.instruments[i].data, &tune[ptr+11], 13);
+        memcpy(bmf.instruments[i].data, tune + ptr + 11, 13);
```
Malvineous, line 177: Same argument here.  Given that `[]` is used a lot elsewhere, I think keeping `[]` in use makes the code a bit more uniform.  Otherwise you could make the argument that on line 161 above, `(tune[ptr] << 24) | (tune[ptr + 1] << 16)` should be replaced with the horrible looking `(*(tune + ptr) << 24) | (*(tune + ptr + 1) << 16))` to match the new change here :)

src/bmf.cpp
```diff
@@ -141,48 +145,46 @@ bool CxadbmfPlayer::xadplayer_load()
...
         ptr += 24;
-	  }
+      }
+      else if (bmf.version == BMF1_1)
+      {
+        memset(bmf.instruments[i].name, 0, 11);
+        memcpy(bmf.instruments[i].data, bmf_default_instrument, 13);
```
Malvineous, line 183: Is there any way you can replace the hard-coded constants with something more future-proof like `sizeof(bmf_default_instrument)`?

src/bmf.cpp
```diff
@@ -141,48 +145,46 @@ bool CxadbmfPlayer::xadplayer_load()
...
-        bmf.instruments[i].name[0] = 0;
-		
-        if (bmf.version == BMF1_1)
-		  for(int j=0;j<13;j++)
-			bmf.instruments[i].data[j] = bmf_default_instrument[j];
-        else
-		  for(int j=0;j<13;j++)
-			bmf.instruments[i].data[j] = 0;
-	  }
+        memset(bmf.instruments + i, 0, sizeof(bmf.instruments[i]));
```
Malvineous, line 186: Yes like `sizeof()` here, this is good.

src/bmf.cpp
```diff
@@ -204,16 +206,17 @@ bool CxadbmfPlayer::xadplayer_load()
...
+
   // load streams
   if (bmf.version > BMF0_9B)
   {
     if (tune_size - ptr < 4)
       return false;
-    unsigned long sflags = (tune[ptr] << 24) | (tune[ptr+1] << 16) | (tune[ptr+2] << 8) | tune[ptr+3];
-    ptr+=4;
+    unsigned long sflags = (tune[ptr] << 24) | (tune[ptr+1] << 16)
+                         | (tune[ptr+2] << 8) | tune[ptr+3];
```
Malvineous, line 216: Does this look neater if rather than shortening it for the sake of line length, you just put each `tune[]` thing on its own line?  Might make it easier to see the pattern?

src/bmf.cpp
```diff
@@ -266,162 +261,159 @@ void CxadbmfPlayer::xadplayer_rewind(int subsong)
...
-        memcpy(&event, &bmf.streams[i][bmf.channel[i].stream_position], sizeof(bmf_event));
+again:
+    const bmf_event &event = bmf.streams[i][pos];
 #ifdef DEBUG
-   AdPlug_LogWrite("%02X %02X %02X %02X %02X %02X\n",
-		   event.note,event.delay,event.volume,event.instrument,
-		   event.cmd,event.cmd_data);
+    AdPlug_LogWrite("%02X %02X %02X %02X %02X %02X\n",
+                    event.note, event.delay, event.volume,
+                    event.instrument, event.cmd, event.cmd_data);
```
Malvineous, line 305: Personally I'm not a fan of excessive indenting without a very good reason, I'd be happy with normal indenting here.

src/bmf.cpp
```diff
@@ -467,193 +459,153 @@ long int CxadbmfPlayer::__bmf_convert_stream(const unsigned char *stream,
...
+  // until the end marker is found and stored.
+  for (int pos = 0; pos <= last_pos; pos < last_pos && pos++)
   {
-    if (pos >= sizeof(bmf.streams[0]) / sizeof(bmf.streams[0][0]))
-    {
-      // Stream too long to store.
-      pos--; // Truncate but keep reading until end marker
-      // or: return -1;
-    }
+    bmf_event &event = bmf.streams[channel][pos];
```
Malvineous, line 475: Always nice to shorten lengthy variable names, I like this.

src/bmf.cpp
```diff
@@ -467,193 +459,153 @@ long int CxadbmfPlayer::__bmf_convert_stream(const unsigned char *stream,
...
       {
-        // byte0: 0aaaaaaa = NOTE
-        bmf.streams[channel][pos].note = *stream & 0x7F;
+        // 2nd byte is 1xDDDDDD: delay byte present
+        event.delay = *stream & 0x3F;
 
-		stream++;
-      } // if (*stream & 0x80)
-    } // if (*stream == 0xFE)
+        if (!(*(stream++) & 0x40))
```
Malvineous, line 526: Please confirm but I believe you can get away without the brackets here and in the other places: `*stream++` (as the `++` doesn't apply until the whole thing has finished being evaluated).

src/bmf.cpp
miller-alex: It's a matter of taste, as I wrote above. While I don't like it very much, it's OK for me to restore the old `&[]` pattern if you prefer that. Just tell me whether you want me to replace the old commits or add a new one.
(Unrelated side note: I'd also prefer memcmp over strncmp here, but it's really minor and some people might not like it either.)

src/bmf.cpp
miller-alex: See above. But to be clear, I don't want to ban `[]`, I just don't like `&[]`. No sane person would write `*(tune + ptr)`, and I've even changed some occurrences of `*(stream+1)` to `stream[1]` further down (although I can see why one might prefer the other form in that particular contex; and now `stream[1]` is used next to `*stream`, which is inconsistent, but `*stream` is a better fit than `stream[0]` there; whatever you do, you have to live with one ugly wart). Some people want conistency, others use different syntax depending on context to emphasize certain aspects, it's all complicated. I don't have a very strong opinion on this matter, just tell me what you want.

src/bmf.cpp
miller-alex: Replacing literal constants with `sizeof` (or an `ARRAY_SIZE()` macro, `enum`, `#define`d constant) is always a good idea. But literal numbers are used everywhere, so I didn't bother. I can change it here if you want, but a general literal constant cleanup pass can (and should) be done sometime later. I'd rather fix some real bugs now.

src/bmf.cpp
miller-alex: Actually this should not be open coded at all. This is a pattern that's used all over the place and many players have their own macros/functions dealing with it in addition to libbinio providing a solution for file data. All that should be consolidated into a common set of helpers used everywhere. But that's out of scope for this PR.

src/bmf.cpp
miller-alex: I think most LogWrites are formatted like this, but I really don't care.

src/bmf.cpp
miller-alex: Yes, the parentheses can be omitted since postfix operators have higher precedence than prefix unary operators, but for some people that's hard to memorize. I like the added clarity of the patentheses, but don't like the cluttered display, so it's a mixed bag, again.

miller-alex:
I don't want to invest too much time into style changes. These are changes that have been sitting in my local repo for a while an I thought I should push them upstream. Originally I mostly wanted to simplify the control flow that was hard to follow, and in preparation I also cleaned up a few other things that I wouldn't have bothered with otherwise.

There are a lot of things that could be cleaned up or changed to a more pleasant style, but I suppose my time is better spent fixing real bugs. I might simplify or refactor some code, but style changes have to wait.

Let me know if/what you want me edit/add, or just change it yourself and merge it, no big deal.

miller-alex:
@Malvineous, I've added patches with most of your suggestions. I think it's ready to merge unless you have more requests.

src/bmf.cpp
Malvineous: I don't mind `strncmp` vs `memcmp` but since we're dealing with binary data here rather than text, `memcmp` seems a bit more logical.  It also gets rid of the typecast so happy with that change.

src/bmf.cpp
Malvineous: Generally my opinion is that `&[]` is clearer to me when talking about offsets in files, as it's obvious you're looking at an offset because of the array syntax.  Something like `tune + 11` requires more mental effort for me, as I have to remember whether `tune` is being used as a pointer or an integer.

src/bmf.cpp
Malvineous: Fair enough, I just figured if you were changing those lines anyway it'd be a good opportunity to fix it while you were there.

src/bmf.cpp
Malvineous: Fair enough!

src/bmf.cpp
Malvineous: You make a fair point, but by the same token memorising the operator precedence is probably going to serve someone well in the long run :)

Malvineous:
All good, thanks for helping to improve the code!

Issue #110, opened 2020-05-12
Multiple additional errors, mostly OOB reads, found by AFL++


debrouxl:
Thanks for integrating Alex Miller's fixes :)
I rebuilt adplug (then at commit 480ed3c9fae57d732b7c12e0afcfd02e3cf4ee88 ) with them and AddressSanitizer instrumentation by AFL++'s afl-clang-fast(++), and restarted fuzzing with AFL++'s afl-fuzz, using the test suite as initial corpus (minus the ref files, but including the .cpp files I didn't notice at first...).
The first crashes (memory errors turned into aborts, segmentation faults, etc.) were found quickly, and a bit over a core-day later, the set of samples which crash asan-instrumented adplug has grown to something I can post. At the time of this writing, afl-fuzz has examined only about a third of the nearly 2000 paths it found, so there's a lot of room for finding more paths and issues.

I ran over half of the samples manually, all of them reproduced. The output shows multiple OOB reads of various sizes, at least one overlapping memcpy, at least one wild read.

[adplug_asan_crashes_20200512_1.tar.gz](https://github.com/adplug/adplug/files/4613620/adplug_asan_crashes_20200512_1.tar.gz)

Malvineous:
@miller-alex Help! :)

miller-alex:
That doesn't surprise me at all. There are a lot of players I haven't looked at yet, but a quick glimpse showed their code quality isn't much better. I had planned to look at the remaining players anyway, but have been busy with other things in the last weeks.
I'll have a look when I have time, but I don't promise anything. And I don't mind if somebody else beats me :)

miller-alex:
@debrouxl I've just added your samples to the tests locally, but I don't see any errors for the following two files, they just refuse to load:
* output/1/crashes/id:000047,sig:06,src:000857,time:11719025,op:havoc,rep:32
* output/2/crashes/id:000033,sig:06,src:001019,time:14254208+000728,op:splice,rep:128
This is probably due to different compiler/sanitizer options. Can you check these samples and post the output, so I can see what's going on?

debrouxl:
My computer has now spent slightly over a core-month CPU time  executing the program 56+M times, finding 2300+ paths and 200 locally unique crashes which still yield "only" 19 unique SUMMARY lines
=> there's been no new crash since 12 days ago...
I guess I'll soon stop the fuzzing jobs and upload the latest set of crashing samples for coverage, but there shouldn't be anything more from that batch to fix.

[...]

PR #113, opened 2020-05-19
Fixes for issue #110

miller-alex:
I just wanted to show some WIP on fixing #110. For now, I've
* added all the fuzzing samples (redundant samples can be removed/disabled later if tests are too slow),
* fixed the crashes in dtm.cpp and cleaned up the file a bit,
* fixed some crashes in a2m.cpp (not finished yet) and did some cleanup/refactoring there.

There's still a lot to do. If somebody wants to help or has comments/suggestions, just add your comments below.


Malvineous:
I'm happy to merge WIP changes, but it would be nice if we could do it in such a way as the tests don't fail in the master branch, as it makes it difficult to pick up other unrelated failures coming in as PRs.  I guess keeping the failing tests in separate branches and then rebase-merging them onto the master branch, along with the fixes when they are ready, would be a way to achieve this.

@miller-alex If this sounds good to you, let me know and I'll give you commit access to the project so you don't have to open PRs all the time.

miller-alex:
My plan was to incrementally add more commits until all tests from the issue are fixed, and then mark the PR as ready. I'm about to push the remaining fixes for a2.cpp and cff.cpp right now.
I've mixed adding test cases, bug fixes, and code cleanup in a single branch this time because it's simpler to work with. If you want to merge the fixes before everything is complete I could move them to a different branch or disable the failing test cases by commenting them out.

I don't mind creating a few pull requests. That gives you or others the chance to review the changes before they land in master. But if you don't want to review them and want me to push changes in smaller batches you can give me direct access. I'm fine either way.

miller-alex:
All problems I've seen with the test samples from issue 110 are fixed, with the following two exceptions.
* A few files write out of range data to OPL registers. That's an old issue and not really a problem since the high bits are simply discarded, so I haven't looked into it.
* Some protracker-based players allow setting tempo=0, and then getrefresh() returns 0. That's no problem for the tests, but other consumers may run into problems with an infinite update interval. (Sample i-110_010.a2m exhibits this issue.) Maybe this should be addressed, but I'm not sure what's the expected behavior (interestingly, in CmodPlayer::update(), command 18 explicitly checks for zero and sets tempo, command 15 sets tempo only for values >= 50, and command 7 doesn't check at all).

I didn't keep separate branches for test, fixes, and cleanup, but the commits will end up in the same master branch anyway. I don't think it makes sense to split them.

From my side that's it for the current batch of fixes. (I can add a commit for the tempo=0 issue if you tell me how that should be handled.) I know there are more problems in the code base, but those have to wait for another PR. @Malvineous I think you can merge this PR now unless you have some change requests.

Malvineous:
Excellent work, thanks so much!

Issue #115, opened 2020-07-08
certain IMF files can't be played


ghost:
Hi

According to the developer of the Adplug plugin for foobar2000, there is a bug in the release version of Adplug which prevents loading some IMF files.

Please read here for all information (and the IMF files that don't load):

https://hydrogenaud.io/index.php?topic=119501.0

Here are the files: http://www.shikadi.net/wiki/files/keenmusic-r3.zip
(look inside "extracted IMFs" folder)

I post this message to ask if this bug is fixed already or still inside Adplug's code.

Thanks
Jessica

jessicajonesunitedkingdom:
Thanks for the fix!

PR #117, opened 2020-07-14
Fix for certain IMF files

emoon:
This is a fix for issue #115 I didn't original do it but comes from this thread https://hydrogenaud.io/index.php?topic=119501.0 and this commit https://g.losno.co/chris/foobar2000/commit/699597fb65ed582df5e6f90209c1fca017f8c860


kvee33 (commented):
Confirmed working, I ran into the same issue and did this exact same fix

PR #130, opened 2021-09-04
Fix various typos

luzpaz:
Found via `codespell -q 3 -L ba,herad,nd,onl,pevents`


luzpaz:
Please review 7519350 closely as it introduces API changes

src/fmopl.c
```diff
@@ -501,7 +501,7 @@ INLINE void OPL_CALC_CH( OPL_CH *CH )
 	}
 }
 
-/* ---------- calcrate rythm block ---------- */
+/* ---------- calcrate rhythm block ---------- */
```
Malvineous, line 504: Should this be "calculate"?

luzpaz: @Malvineous `calcrate`appears in other parts of the code...so I wasn't sure if it was referencing this internal jargon

luzpaz: @Malvineous can you confirm?

src/fmopl.c
```diff
@@ -657,7 +657,7 @@ static int OPLOpenTable( void )
 		TL_TABLE[t] = TL_TABLE[TL_MAX+t] = 0;
 	}
 
-	/* make sinwave table (total level offet) */
+	/* make sinwave table (total level offset) */
```
Malvineous, line 660: Just noticed this should really be "sine wave" instead of "sinwave" too

src/fmopl.c
Malvineous: It's hard to say but I can't think what other word it could be.  I think change them all to "calculate", unless you can find any that have nothing to do with mathematical operations.

src/fmopl.c
luzpaz: Added requested change in 07c832c

Malvineous:
Rebased and merged in 6cfb13795eb511d64bc2dcb7caf063fe02798934

Issue #40, opened 2016-10-12
Add support for type-1 MIDI files


ghost:
I know it was reported elsewhere a long time ago, but it isn't on this issue tracker yet so I thought I'd contribute. MIDI playback on Type 1 MIDI files does not work. The duration is correctly detected, but no sound comes out. I'm a Winamp user.

I'd love to be able to play back the Duke Nukem 3D MIDI files in ADPlug and reminisce of the old days!

PR #132, opened 2022-02-06
Detect type 1 (multi-track) MIDI files.

mywave82:
Also increases the base-octave one up, since all test files (Duke Nukem 3D and Doom) ended up the very low bass area.

Fixes bug adplug #40


src/mid.cpp
```diff
@@ -438,6 +449,8 @@ void CmidPlayer::midi_fm_playnote(int voice, int note, int volume)
     int oct=note/12;
 	int c;
 
+    midiprintf ("playnote(voice %d, note %d, volume %d)\n", voice, note, volume);
```
Malvineous, line 452: Can these (and the other) debug statements be removed?  It looks like they are left over from your testing.

src/mid.cpp
```diff
@@ -941,21 +955,50 @@ void CmidPlayer::rewind(int subsong)
                 adlib_style=LUCAS_STYLE|MIDI_STYLE;
                 //note: no break, we go right into midi headers...
             case FILE_MIDI:
+            case FILE_MIDItype1:
                 if (type != FILE_LUCAS)
                     tins=128;
-                getnext(11);  /*skip header*/
+                /* "Mthd"              header-chunk
```
Malvineous, line 961: Should be `MThd`

src/mid.cpp
```diff
@@ -941,21 +955,50 @@ void CmidPlayer::rewind(int subsong)
...
+                type=(i==1)?FILE_MIDItype1:FILE_MIDI;
+		midiprintf ("type:%ld\n", i);
+                getnext(2); /* skip track-number */
                 deltas=getnext(2);
                 midiprintf ("deltas:%ld\n",deltas);
-                getnext(4);
+
+	        for (i=0; i<16; i++)
+	        {
+                  ch[i].nshift=-13;
```
Malvineous, line 977: Is this where it does the octave shift?  Could you add a comment to say why this is done?  It just seems like it might be the wrong place to fix the problem, so someone might want to remove it later if they work out where the proper fix should be.

src/mid.cpp
```diff
@@ -1129,6 +1172,8 @@ std::string CmidPlayer::gettype()
 		return std::string("LucasArts AdLib MIDI");
 	case FILE_MIDI:
 		return std::string("General MIDI");
```
Malvineous, line 1174: Could you mention this is type-0 as well so it's clear what the difference is?

src/mid.cpp
```diff
@@ -114,6 +114,7 @@ typedef unsigned int   uint32_t;
 // File types
 #define FILE_LUCAS      1
 #define FILE_MIDI       2
+#define FILE_MIDItype1  7 // Multiple track file format
```
Malvineous, line 117: Can you update `FILE_MIDI` to be `FILE_MIDItype0` as well so it's clear what the difference is?

src/mid.cpp
```diff
@@ -941,21 +955,50 @@ void CmidPlayer::rewind(int subsong)
...
+                while (((type==FILE_MIDI) && curtrack == 0) ||
+                       ((type==FILE_MIDItype1) && (curtrack < 16)))
+                {
+                    char s[4];
+                    s[0] = 0;
+                    readString(s, 4);
+                    midiprintf("Offset=0x%08lx\n", pos);
+                    midiprintf("trkHeader: %c%c%c%c\n", s[0], s[1], s[2], s[3]);
+                    if (s[0] != 'M' || s[1]!='T' || s[2]!='r' || s[3]!='k')
+                        break;
```
Malvineous, line 991: It might be neater here to make `s` longer (`char s[5]`) then after `readString()` you can put `s[4] = 0;` to ensure it is a null-terminated string.  Then `s` will be a normal string so you can say `if (strcmp(s, "MTrk") == 0)` and likewise `trkHeader: %s`.

Malvineous (dismissed):

src/mid.cpp
mywave82: Good catch :-)

src/mid.cpp
mywave82: (The shift also included one semi-tone). I will try to find some midi files that makes it easier to verify if this is correct or not.

src/mid.cpp
mywave82: Please note that it does not detect if it is type0 or type1 until rewind() has been called if I remember the code correct, since it needs to parse more header data than is done in the initial detection routine.

src/mid.cpp
mywave82: Good idea. I was just following the coding-standard already used in this file.

src/mid.cpp
mywave82: (courtesy of https://github.com/cirlabs/miditime for generating test midifile)
[miditest.zip](https://github.com/adplug/adplug/files/8103942/miditest.zip)

The very first note should be a C-4 at 261Hz, while adplug plays it back at 129Hz which is slightly below a C-3.

The -13 is because the default value in the midi playback engine is -25, so increasing one octave == +12 semitones, that gives us -13 instead.

mywave82: That leaves Lucas MIDI files - they must be tested if they should have -25 or -13. They will currently have -13. I do not have the files, nor I do I know how they should sound. Can you test if they are too high now, if so I will add an if() statement

binarymaster: This change does increase all notes by one octave in the existing music test file `mi2.laa`, so the test fails. I don't know if this makes it sound more correct, needs to be tested with original player or game.

I think it's better to propose this fix in a separate PR, since it's not really related to detecting MIDI type 1 files.

src/mid.cpp
binarymaster: I have added a new member to the class for this, as the code looks more clear this way.

src/mid.cpp
```diff
@@ -943,19 +966,38 @@ void CmidPlayer::rewind(int subsong)
...
                     tins=128;
-                getnext(11);  /*skip header*/
+                getnext(11); /* skip past header data until deltas is reached */
                 deltas=getnext(2);
                 midiprintf ("deltas:%ld\n",deltas);
-                getnext(4);
+	        for (i=0; i<16; i++)
+	        {
+                  ch[i].nshift=-13;
+                  ch[i].on=1;
```
binarymaster, line 975: Also this change is not clear for me, why we need to mark all channels as "on"?

I have tested with several MIDI files, it doesn't have any audible effect to me.

src/mid.cpp
```diff
@@ -309,7 +319,20 @@ bool CmidPlayer::load(const std::string &filename, const CFileProvider &fp)
...
+                 * 0x00 0x00 0x00 0x06 header size
+                 * 0x00 0x0n           midi file type: 0=single-track format, 1=multiple-track format, 2=multiple-song format
+                 * 0xnn 0xnn           track-number
+                 * 0xnn 0xnn           tempo
+                 */
+                unsigned char s2[4];
+                f->readString((char *)s2, 4);
+                midi_type=u16_unaligned(s2 + 2);
+                good=FILE_MIDI;
+                midiprintf ("General MIDI type: %d\n", midi_type);
```
binarymaster, line 334: Reading MIDI type-1 this way results in variable getting wrong value (256 instead of 1), because the value is stored in Big Endian format. Better rewrite this part with binio functions.

binarymaster:
> Also increases the base-octave one up, since all test files (Duke Nukem 3D and Doom) ended up the very low bass area.

Yes you're right, I can confirm this, but since it breaks LAA playback, it needs to be investigated separately.

src/mid.cpp
mywave82: MIDI files are without this at all one semi-tone off. How much is LAA files off, without this patch applied. Perfect or one semi-tone off?

src/mid.cpp
```diff
@@ -943,19 +970,33 @@ void CmidPlayer::rewind(int subsong)
...
-                midiprintf ("tracklen:%lu\n",track[curtrack].tend);
-                //track[curtrack].tend += pos; // FIXME: length -> end position
-                if (track[curtrack].tend > flen) // no music after end of file
-                    track[curtrack].tend = flen;
-                track[curtrack].spos=pos;
+                while ((curtrack == 0) ||
+                       ((midi_type == 1) && (curtrack < 16)))
+                {
+                    /* MIDI type 0 (and LucasArts AdLib MIDI) stores all the MIDI channels in a single track,
+                     * while MIDI type 1 splits each channel into separate tracks */
```
binarymaster, line 982: Well actually MIDI type-1 may have more than 16 tracks, and some of them may share one channel, like drum tracks (one track per drum type, but they all use channel 10 anyway).

src/mid.cpp
binarymaster: I don't have any LAA files to check, judging only by one single file in the tests.

binarymaster: Please file a separate issue regarding MIDI files having one octave + one semi-tone off.

binarymaster (changes requested):
I have tested this PR, here are things to do:
- [ ] MIDI type-1 files are detected, but only track 0 is played
- [ ] Add some test music files and their references

Yes, and in case of `GRABBAG.MID` from Duke3D, it plays only the bass line. :smile:

binarymaster (approved):

PR #133, opened 2022-04-18
Add missing RAD2 test files to EXTRA_DIST

atsampson:
Since 09b4a007a33de41b8eb5ce1eb491687f43c263be, playertest needs these two files. However, they weren't listed in EXTRA_DIST, so they were missing from the 2.3.3 tarball, which means running the test suite from the tarball fails.


Malvineous:
Many thanks!

PR #134, opened 2022-04-30
Fix compiler warnings (compiler used is g++ v11.2.0)

mywave82:
There are also quite many "comparison of integer expressions of different signedness" that I have not touched.

Functionality has not changed


binarymaster:
This has broken the build:
```sh
src\adl.cpp(2918): error C3861: 'snprintf': identifier not found 
```

binarymaster:
Fixed build in e8555ce1.

PR #135, opened 2022-08-01
Fix a crash with RAD2 files

kode54:
Affected files: xmasong.rad, possibly others

Song references unused instruments not included in the file, causing uninitialized instrument Riff member to be dereferenced, which ends up pointing at unallocated memory.


binarymaster (approved):
Sorry for long delay, the change looks good to me. :+1:

PR #137, opened 2022-11-27
Split off the synth class from ROL player and use it

binarymaster:
It took me 3 days to do this, now it should be polished and ready for merging. Each commit is separately buildable and passes tests. I will merge it in a few days, but it would be nice to get some review.

The main point is code deduplication. Less code - less size and faster building. Also I'm not certain about adlib.c/adlib.h license (I wasn't thinking throughtly while adding it at first), so this makes it even better.

The second point, previous BNK loading code was slow and unoptimized. Since MUS/IMS player now uses ROL backend code to load BNK files, now it does that much more efficiently.

Also Cc @OPLx and @mywave82, as I cannot add them to reviewers.


Malvineous (approved):
Nice work!  The code looks fine to me and as long as the tests pass that's the main requirement!

mywave82:
It is good logical step :-)

PR #140, opened 2022-12-11
Include debug.h in coktel.cpp.

henricj:
The file `coktel.cpp` does two log writes (on line [214](https://github.com/adplug/adplug/blob/ba20f6e4405590974686d3fbea2a7abdabb5b511/src/coktel.cpp#L214) and [222](https://github.com/adplug/adplug/blob/ba20f6e4405590974686d3fbea2a7abdabb5b511/src/coktel.cpp#L222)), but does not include `debug.h`.

### DIFF
--- a/src/coktel.cpp
+++ b/src/coktel.cpp
@@ -27,6 +27,7 @@
 #include <string.h>
 
 #include "coktel.h"
+#include "debug.h"
 
 /*** public methods *************************************/
 


src/coktel.cpp
```diff
@@ -27,6 +27,7 @@
 #include <string.h>
 
 #include "coktel.h"
+#include "debug.h"
```
binarymaster, line 30: Minor suggestion, include it only when DEBUG is defined.
```suggestion

#ifdef DEBUG
#include "debug.h"
#endif
```

binarymaster (approved):
Indeed I have missed that, nice catch!

PR #144, opened 2022-12-12
Fix for a2m -Wsign-compare

mywave82:
src/a2m.cpp: In member function ‘virtual bool Ca2mLoader::load(const string&, const CFileProvider&)’: src/a2m.cpp:162:17: warning: comparison of integer expressions of different signedness: ‘int’ and ‘long unsigned int’ [-Wsign-compare]
  162 |   for (i = 0; i < length; i++)
src/a2m.cpp: In member function ‘virtual bool Ca2mLoader::load(const string&, const CFileProvider&)’:
src/a2m.cpp:203:20: warning: comparison of integer expressions of different signedness: ‘long int’ and ‘size_t’ {aka ‘long unsigned int’} [-Wsign-compare]
  203 |   if (orgptr - org < needed) {

### DIFF
--- a/src/a2m.cpp
+++ b/src/a2m.cpp
@@ -47,7 +47,8 @@ CPlayer *Ca2mLoader::factory(Copl *newopl)
 bool Ca2mLoader::load(const std::string &filename, const CFileProvider &fp)
 {
   binistream *f = fp.open(filename); if (!f) return false;
-  int i, j, k, t;
+  unsigned int i;
+  int j, k, t;
   unsigned int l;
   unsigned char *org, *orgptr, flags = 0;
   unsigned long alength;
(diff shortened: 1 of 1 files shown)


mywave82:
There are more of these signed compare warnings, but probably makes sense to split them up into PRs, since some of them might need discussion? Or do you prefer to have them in a single PR?

src/a2m.cpp
```diff
@@ -199,7 +200,7 @@ bool Ca2mLoader::load(const std::string &filename, const CFileProvider &fp)
     orgptr = org + alength;
   }
 
-  if (orgptr - org < needed) {
+  if (orgptr - org < (ssize_t)needed) {
```
mywave82, line 203: pointer - pointer   =>  signed long
_needed_ in most locations are used as an unsigned number except this single comparison

binarymaster: From MSVC: `error C2065: 'ssize_t': undeclared identifier`

binarymaster:
> There are more of these signed compare warnings, but probably makes sense to split them up into PRs, since some of them might need discussion?

Separate PRs for each would probably be better, because easier to review and merge.

PR #148, opened 2022-12-17
Fix for d00-Wsign-compare

mywave82:
```
src/d00.cpp: In member function ‘virtual void Cd00Player::rewind(int)’: src/d00.cpp:465:15: warning: comparison of integer expressions of different signedness: ‘int’ and ‘unsigned int’ [-Wsign-compare]
  465 |   if (subsong < getsubsongs() && dataofs + sizeof(Stpoin) <= filesize)
      |       ~~~~~~~~^~~~~~~~~~~~~~~
```

### DIFF
--- a/src/d00.cpp
+++ b/src/d00.cpp
@@ -462,7 +462,7 @@ void Cd00Player::rewind(int subsong)
 
   size_t dataofs = subsong * sizeof(Stpoin)
     + LE_WORD(&(version > 1 ? header->tpoin : header1->tpoin));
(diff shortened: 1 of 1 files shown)


mywave82:
Currently rewind() is prototyped with subsong being signed int, and only rix.cpp and d00.cpp actually has special handling for <0 value, while other subsong implementors does range checks and clamps the value to 0 if out of range. I assume the correct would be to fix other implementors to handle <0 as current song reload instead of clamping.

mywave82:
Moved rewind() question into is own issue.

PR #149, opened 2022-12-17
Fix for dro2-Wsign-compare and convert some additional variables from signed into unsigned 

mywave82:
Fix this warning and make some more of the logic work with unsigned instead of signed datatypes.

```
src/dro2.cpp: In member function ‘virtual bool Cdro2Player::load(const string&, const CFileProvider&)’: src/dro2.cpp:69:27: warning: comparison of integer expressions of different signedness: ‘int’ and ‘long unsigned int’ [-Wsign-compare]
   69 |             this->iLength > fp.filesize(f) - f->pos()) {
      |             ~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/src/dro2.cpp
+++ b/src/dro2.cpp
@@ -64,8 +64,8 @@ bool Cdro2Player::load(const std::string &filename, const CFileProvider &fp)
 		return false;
 	}
 
-	this->iLength = f->readInt(4); // should better use an unsigned type
-	if (this->iLength <= 0 || this->iLength >= 1<<30 ||
+	this->iLength = f->readInt(4);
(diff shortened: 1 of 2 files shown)


src/dro2.cpp
```diff
@@ -64,8 +64,8 @@ bool Cdro2Player::load(const std::string &filename, const CFileProvider &fp)
 		return false;
 	}
 
-	this->iLength = f->readInt(4); // should better use an unsigned type
```
mywave82, line 67: // should better use an unsigned type

Acknowledged

PR #156, opened 2022-12-18
Fix for psi-Wsign-compare 

mywave82:
Fix these warnings:

```
src/psi.cpp: In member function ‘virtual bool CxadpsiPlayer::xadplayer_load()’: src/psi.cpp:60:44: warning: comparison of integer expressions of different signedness: ‘int’ and ‘long unsigned int’ [-Wsign-compare]
   60 |   if (header.instr_ptr + psi.nchannels * 2 >= tune_size ||
      |       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~
src/psi.cpp:61:42: warning: comparison of integer expressions of different signedness: ‘int’ and ‘long unsigned int’ [-Wsign-compare]
   61 |       header.seq_ptr + psi.nchannels * 4 >= tune_size)
      |       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~
src/psi.cpp:70:40: warning: comparison of integer expressions of different signedness: ‘int’ and ‘long unsigned int’ [-Wsign-compare]
   70 |     if (le16(&psi.instr_table[i]) + 11 >= tune_size)
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~
```

### DIFF
--- a/src/psi.cpp
+++ b/src/psi.cpp
@@ -57,8 +57,8 @@ bool CxadpsiPlayer::xadplayer_load()
   header.instr_ptr = le16(&tune[0]);
   header.seq_ptr = le16(&tune[2]);
 
-  if (header.instr_ptr + psi.nchannels * 2 >= tune_size ||
-      header.seq_ptr + psi.nchannels * 4 >= tune_size)
+  if (header.instr_ptr + (unsigned)psi.nchannels * 2 >= tune_size ||
+      header.seq_ptr + (unsigned)psi.nchannels * 4 >= tune_size)
     return false;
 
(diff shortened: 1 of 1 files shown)


mywave82:
Challenge is that psi.nchannels is an enum, and it is used to define the member-sizes in the same structure. It will always be 8, unless the struct is made into a class. Perhaps a better solution is to make a define instead:

`#define PSI_NCHANNELS 8`

PR #162, opened 2022-12-30
Add .editorconfig

henricj:
This adds an [EditorConfig](https://editorconfig.org/) file that lets many common code editors know how to indent.


mywave82:
indentation varies from file to file in this project, since they have different origins which makes it tricky

PR #176, opened 2023-03-20
mid.cpp: Replace std::to_string with compatible solution

dmitrysmagin:
Trivia: older mingw/gcc 4.8.x versions don't have std::to_string implementation.
And since to_string is used only in one sole case in mid.cpp to show the midi_type value, it can be safely replaced with compatible solution.
'midi_type' can hold 0, 1 or 2 only, so we should care only about these values.

Why bother? There are still some cross-toolchains that are nailed to old gcc versions that won't be ever updated. The rest of the code is absolutely 4.8.x compliant, so why not make it slightly more compatible?

### DIFF
--- a/src/mid.cpp
+++ b/src/mid.cpp
@@ -1170,7 +1170,7 @@ std::string CmidPlayer::gettype()
 	case FILE_LUCAS:
 		return std::string("LucasArts AdLib MIDI");
 	case FILE_MIDI:
-		return std::string("General MIDI (type " + std::to_string(midi_type) + ")");
+		return std::string("General MIDI (type " + std::string(1, midi_type | 0x30) + ")");
 	case FILE_CMF:
 		return std::string("Creative Music Format (CMF MIDI)");
 	case FILE_OLDLUCAS:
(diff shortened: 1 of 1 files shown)


mywave82:
last alternative would be to have a sub switch with different text?
"type 0 - single track"
"type 1 - multi track"
"type 2 - multi pattern"

dmitrysmagin:
I think, showing just 'Type 0' or 'Type 1' would be enough. Any other objections?

sagamusix:
Maybe a compromise that doesn't assume ASCII but also doesn't pull in `stringstream` would be `std::string(1, '0' + midi_type)` which doesn't hardcode '0' as 0x30 and thus is also a bit more understandable.

BTW, the readme still mentions that AdPlug is "currently tested" with GCC 4.1, which evidently cannot be true. Maybe that information should be removed...

mywave82:
tweak from @sagamusix is more readable, and with a comment perhaps about why we do it.
```c++
return std::string("General MIDI (type " + std::string(1, '0' + midi_type) + ")"); // avoid using std::to_string(midi_type) which is c++11 ```

I also added a issue about that testing currently does not test gcc 4.1, this detail probably has been missing since adplug was migrated to github

dmitrysmagin:
I've updated the PR as @mywave82 suggested.

mywave82:
Thank you for the patience

PR #185, opened 2023-04-26
Loader for .mtr tunes (Master Tracker)

dmitrysmagin:
Hi.
I've noted that mtr is still not supported: http://adplug.github.io/library/entry/ArkhamMasterTracker24.html
so I've added a loader which reuses protrack.cpp.

Notes:
- Original AKMTRK2.DOC has lots of errors and inconsistencies on file format, so some guess was applied.
- Almost all commands are translated except for retrigger and order jumps, but I didn't find any tunes that use them.
- Most tunes sound very close compared with the original tracker, but some tunes (especially converted from MODs) have ridiculous out-of-tune slides. That's not a player's error, the tunes are composed like that.


mywave82:
I will start looking at this PR sometime after the weekend

src/mtr.cpp
```diff
@@ -0,0 +1,211 @@
...
+    char header[51] = {0};
+    char mtitle[21] = {0};
+    int nvoices, ndigvoices, npatterns, orderlen,
+        timervalue = 0x428F, restart, flength;
+
+    f->readString(header, 50);
+
+    if (!strncmp(header, "MTRAC ", 6)) {
+        version = 1;
+        sscanf(header + 26, "%02x %02x %02x %02x %08x",
```
mywave82, line 50: Need to error-out if sscan() returns another value than 5

(Which will be the case for .MTR (v1.x) files)

src/mtr.cpp
```diff
@@ -0,0 +1,211 @@
...
+            &orderlen,
+            &restart,
+            &flength
+        );
+        strncpy(mtitle, header + 6, 20);
+        timervalue = f->readInt(2); // usually 428F
+        f->ignore(1); // 0=SPK 1=ADL 2=SBP
+    } else if (!strncmp(header, "MTRACK NC", 9)) {
+        version = 2;
+        sscanf(header + 10, "%02x %02x %02x %02x %02x %02x %04x %08x",
```
mywave82, line 62: Need to error-out if sscan() returns another value than 8

mywave82 (changes requested):

src/mtr.cpp
```diff
@@ -0,0 +1,211 @@
...
+    ninstruments = 0;
+
+    char header[51] = {0};
+    char mtitle[21] = {0};
+    int nvoices, ndigvoices, npatterns, orderlen,
+        timervalue = 0x428F, restart, flength;
+
+    f->readString(header, 50);
+
+    if (!strncmp(header, "MTRAC ", 6)) {
```
mywave82, line 48: Documentation says "MTRACK ". Is the documentation incorrect, if so, add a comment please

mywave82 (changes requested):

src/mtr.cpp
```diff
@@ -0,0 +1,211 @@
...
+    } else if (!strncmp(header, "MTRACK NC", 9)) {
+        version = 2;
+        sscanf(header + 10, "%02x %02x %02x %02x %02x %02x %04x %08x",
+            &nvoices,
+            &ndigvoices, // unused
+            &npatterns,
+            &orderlen,
+            &ninstruments,
+            &restart,
+            &timervalue, // usually 428F
```
mywave82, line 69: timervalue also seem to be a deviation from the documentation, please add comment :-)

mywave82 (changes requested):

src/mtr.cpp
dmitrysmagin: v1 files have "MTRACK\x20" at the beginning, v2 have "MTRACK\x20NC\x20".
sscanf shouldn't fail once the version is recognized.
AKMTRK2.DOC is incorrect describing this. There are lots of errors there, too much to enumerate them all, easier to create another 'fixed' variant of a doc.

src/mtr.cpp
dmitrysmagin: It is incorrect.

src/mtr.cpp
dmitrysmagin: Yep, in v1 it's a binary value after the string header, in v2 it's one of the hex values in the header string. Obviously, this is a timer value written into timer ports, so hz is recalculated back as 0x1234DD / timervalue

src/mtr.cpp
dmitrysmagin: Okay, some error handling for sscanf won't hurt if the file is severely broken.

src/mtr.cpp
mywave82: We might get lucky to find a user with a rare file from a version that is not in the modland.com archives

dmitrysmagin:
Added error checking on sscanf().
I've tested all .MTR files from the archive with the Master Tracker, all seem to be recognized fine.
If there are others somewhere, I'll be willing to test them.

src/mtr.cpp
dmitrysmagin: Fixed.

src/mtr.cpp
dmitrysmagin: I've checked http://modland.com/pub/modules/Ad%20Lib/Master%20Tracker/ and it contains the same MTR modules that were packed with the original Master Tracker. I've tested them all with this loader, and they play identically with MT.

mywave82:
Thank you for your work and patience

PR #193, opened 2023-09-02
Add DJGPP support to Adplug

AranVink:
Add a few #ifdef preprocessor commands to check for DJGPP being used a compiler for DOS.
Prerequisite PR for PR on Adplay-DOS: https://github.com/adplug/adplay-dos/pull/3

### DIFF
--- a/adplugdb/getopt.c
+++ b/adplugdb/getopt.c
@@ -101,7 +101,8 @@
    GNU application programs can use a third alternative mode in which
    they can distinguish the relative order of options and other arguments.  */
 
-#include "getopt.h"
+/* If this file is compiled, it is because our target system is lacking getopt.h */
+#include "mygetopt.h"
 
 /* For communication from `getopt' to the caller.
    When `getopt' finds an option that takes an argument,
--- a/src/rix.cpp
+++ b/src/rix.cpp
@@ -159,7 +159,7 @@ void CrixPlayer::rewind(int subsong)
   data_initial();
 }
 
-uint32_t CrixPlayer::getsubsongs()
+unsigned int CrixPlayer::getsubsongs()
 {
 	if(flag_mkf)
 	{
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,6 +15,10 @@ jobs:
         include:
           - compiler: gcc-4.8 # To test compatibility with older systems
             os: ubuntu-latest
+          - compiler: djgpp-2.0.5-gcc-12.2.0 # To test compatibility for Adplay - DOS
+            os: ubuntu-latest
+          - compiler: djgpp-2.0.5-gcc-4.8.5 # To test compatibility for Adplay - DOS
+            os: ubuntu-latest
 
       fail-fast: false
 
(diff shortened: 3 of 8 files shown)


src/rix.cpp
```diff
@@ -159,7 +159,12 @@ void CrixPlayer::rewind(int subsong)
   data_initial();
 }
 
+#ifdef DJGPP
```
mywave82, line 162: looking at player.h, it appears that it should be `unsigned int getsubsongs()` for all systems, not just DJGPP

src/player.h
```diff
@@ -22,7 +22,11 @@
 #ifndef H_ADPLUG_PLAYER
 #define H_ADPLUG_PLAYER
 
+#if not defined DJGPP
```
mywave82, line 25: I see some mentions online about a bug in DJGPP, that you have to ensure that #include <string> happens before any other header files that uses namespaces.

Is this due to this?

adplugdb/getopt.c
```diff
@@ -101,7 +101,12 @@
    GNU application programs can use a third alternative mode in which
    they can distinguish the relative order of options and other arguments.  */
 
+#ifdef DJGPP
+#include "mygetopt.h"
```
mywave82, line 105: Is mygetopt.h provided by DJGPP, or is this file missing in the PR?

AranVink: Already part of Adplug, it's in the same folder, https://github.com/adplug/adplug/blob/master/adplugdb/mygetopt.h

mywave82: My mistake, I looked in the wrong directory

AranVink:
After some fiddling with GitHub actions I managed to add DJGPP compilation support.
It can now use DJGPP - GCC 12.2.0 and DJGPP - GCC 4.8.5. To get GCC 4.8.5 to work I had to include stdint.h and remove some older specific redeclarations of uint*.
Running tests is really hard to do AFAIK when cross-compiling like this, so it only checks if it can run `make all`. Maybe when I've updated Adplay for DOS I could look into some way of adding tests there. (Also see https://github.com/adplug/adplay-dos/pull/3)

Please review and let me know what you think. I'm very sorry for the commit spam, I thought I was only pushing to my own repository while doing these tests, but unfortunately this PR and CI still got triggered through my own repo/branch. 😅

mywave82:
I am quite busy the next two weeks, but I will try to grab some time every now and and probably rebuild to commit history (doing some git rebase magic)

AranVink:
Much appreciated, if you want me to make a new PR with all commits squashed let me know (I made this mess 😉)

mywave82:
I rebased your tree, so unless you are very familier with git, just do a fresh clone of your own git repository on your computer.

I believe a lot of the fiddling with setting CXXFLAGS etc. to include -I/foo/djgpp/include can be avoided if djgpp extracted into /usr/local/djgpp, as that is probably the PREFIX is was compiled for.

binarymaster:
I think some fixes from here could be already pushed to master, like header and type fixes. They are trivial.

mywave82:
We are getting close to something now (git rebase is needed again to clean up the history)

mywave82:
Sorry for the extra spam lately, trying to make the github workflow be a smooth as possible.

AranVink:
Thank you for following up and making the tweaks so it can be merged, again much appreciated. I'll continue working on Adplay for DOS, building a CD pipeline, and maybe even a CI pipeline (when tooling allows).

PR #194, opened 2023-10-22
WoodyOPL: remove global static variables.

mywave82:
Make them either static or part of the class depending on the usage.

### DIFF
--- a/src/woodyopl.cpp
+++ b/src/woodyopl.cpp
@@ -6,12 +6,12 @@
  *  modify it under the terms of the GNU Lesser General Public
  *  License as published by the Free Software Foundation; either
  *  version 2.1 of the License, or (at your option) any later version.
- * 
+ *
  *  This library is distributed in the hope that it will be useful,
  *  but WITHOUT ANY WARRANTY; without even the implied warranty of
  *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
  *  Lesser General Public License for more details.
- * 
+ *
  *  You should have received a copy of the GNU Lesser General Public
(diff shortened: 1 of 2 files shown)


mywave82:
gcc 13.1.0 gave warnings about assigning local arrays into global variables.

So I decided to move all the global statics that could not be made const to inside the class.

mywave82:
Should probably change .SO version, since the CWemuopl contains a instance of OPLChipClass that is not allocated

binarymaster:
Is this in sync with upstream? (if not, any chance to send patch upstream?)

mywave82:
Upstream for this file will be DosBox (vanilla), and we can push this upstream to them when this patch is finished here.

mywave82:
Upstream DOSBOX and DOSBOX-X are not using classes, so upstream would require more changes. We can notify them after we apply.

Malvineous:
Yeah the reason why this is "messy" is because it's almost a direct copy and paste from the DOSBox source code.  This was done to make it easier to update as changes are made to the DOSBox version - we can just copy their file here.

Once we start making changes to this file the two codebases will diverge, and it will be harder to bring updates across from the upstream code.

So I'm in two minds about these sorts of changes.  Yes it's good to clean up the code, but it will then make it much harder to bring across any changes from the upstream codebase.

mywave82:
The original DOSBox project are only doing bugfixes and compatibility-fixes.
DOSBox-X seems to very quite active, but not on this code.

The goal is to locate all the parts that potentially are using global variables, and thus multithread-safe and ensure that behaviour is consistent.

Not all of these PRs has to be accepted ;-)

PR #196, opened 2023-10-22
[RIX] Make for40reg a regular member, so each instance of RIX playback will have its own working-copy.

mywave82:


binarymaster:
```cpp
  memset(for40reg,   0x7F,  sizeof(for40reg)); // FIXME: !static
```
This FIXME comment from `rix.cpp` could be then removed.

mywave82:
> ```c++
>   memset(for40reg,   0x7F,  sizeof(for40reg)); // FIXME: !static
> ```
> 
> This FIXME comment from `rix.cpp` could be then removed.

Thanks, saw it, but somehow forgot.

binarymaster (approved):

PR #199, opened 2023-12-21
IMS: add .kpf as alternate extension

mistydemeo:
This is used in Family Production's Shakii: https://moddingwiki.shikadi.net/wiki/Shakii_The_Wolf

A few other Family Production games use different file extensions. Should I add those too?


Malvineous:
I think if a game uses a standard format but a custom file extension we should probably not add it, because so many games do this.  For example Jill of the Jungle renames .cmf files to .ddt, Kiloblaster renames .cmf to .kb1, .kb2, .kb3, etc.

If we added all those I think we'd end up with too many obscure extensions and then risk conflicting with other file formats.  In most of these games, the process of "extracting" the music from the game involves renaming the files back to their official extension, so that means we don't need to support all the custom ones.

Do you think we could make that argument here?  That rather than playing the game's files directly, they should be "extracted" (renamed) and then played with standard filename extensions?

mistydemeo:
Makes sense to me. I can just copy them out of the game directory and rename them to match adplug's expectations.

Issue #143, opened 2022-12-12
Issues reported by Coverity


henricj:
A [GitHub workflow runs Coverity](https://github.com/henricj/dunelegacy/blob/modernize/.github/workflows/coverity-scan-ubuntu.yml) for another project that uses adplug as a submodule.  These were reported as new defects yesterday (I have not reviewed them):

```
Hi,

Please find the latest report on new defect(s) introduced to henricj/dunelegacy found with Coverity Scan.

4 new defect(s) introduced to henricj/dunelegacy found with Coverity Scan.
2 defect(s), reported by Coverity Scan earlier, were marked fixed in the recent build analyzed by Coverity Scan.

New defect(s) Reported-by: Coverity Scan
Showing 4 of 4 defect(s)


** CID 300174:  Uninitialized members  (UNINIT_CTOR)
/home/runner/work/dunelegacy/dunelegacy/external/adplug/adplug/src/coktel.h: 53 in CcoktelPlayer::CcoktelPlayer(Copl *)()


________________________________________________________________________________________________________
*** CID 300174:  Uninitialized members  (UNINIT_CTOR)
/home/runner/work/dunelegacy/dunelegacy/external/adplug/adplug/src/coktel.h: 53 in CcoktelPlayer::CcoktelPlayer(Copl *)()
47     {
48     public:
49      static CPlayer *factory(Copl *newopl);
50
51      CcoktelPlayer(Copl *newopl)
52              : CcomposerBackend(newopl), insts(0), data(0)
>>>     CID 300174:  Uninitialized members  (UNINIT_CTOR)
>>>     Non-static class member "modifyTimbre" is not initialized in this constructor nor in any functions that it calls.
53              { }
54      ~CcoktelPlayer()
55      {
56              if (insts) delete [] insts;
57              if (data) delete [] data;
58      };

** CID 300173:  Security best practices violations  (STRING_OVERFLOW)
/home/runner/work/dunelegacy/dunelegacy/external/adplug/adplug/src/composer.cpp: 520 in CcomposerBackend::load_bnk_instrument(binistream *, const CcomposerBackend::SBnkHeader &, const std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> &)()


________________________________________________________________________________________________________
*** CID 300173:  Security best practices violations  (STRING_OVERFLOW)
[...]

PR #204, opened 2024-01-28
Minor "fixes" to coktel/composer to ensure safety:

mywave82:
* Add some assertions to assure that `voice` does not go beyond limits when mRhythmMode is not set
 * Initialize all variables in the constructor, even if they will be set to known values during rewind()
 * Use strncpy() instead of strcpy()
 
 This is a response to #143 

@binarymaster

### DIFF
--- a/src/composer.cpp
+++ b/src/composer.cpp
@@ -150,6 +150,7 @@ CcomposerBackend::CcomposerBackend(Copl * const pNewOpl)
     , mNoteCache         (kNumPercussiveVoices, 0)
     , mKOnOctFNumCache   (kNumMelodicVoices, 0)
     , mKeyOnCache        (kNumPercussiveVoices, false)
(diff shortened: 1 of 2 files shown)


src/composer.cpp
```diff
@@ -246,6 +247,11 @@ void CcomposerBackend::SetNotePercussive(int const voice, int const note)
 //---------------------------------------------------------
 void CcomposerBackend::SetNoteMelodic(int const voice, int const note)
 {
+    if (voice >= 9)
```
binarymaster, line 250: Is it possible to use `kNumMelodicVoices` class constant? (and in other places)
```suggestion
    if (voice >= kNumMelodicVoices)
```

mywave82: Sure, looks more pretty

binarymaster (approved):

PR #207, opened 2024-02-26
Add player for A2M (versions 1-14) and A2T (versions 1-14)

dmitrysmagin:
Hi.
After years of converting FreePascal to C I was able to commit this code to adplug...

This is a direct translation of the player routine from the sources of the Adlib Tracker 2 (http://adlibtracker.net/). A2M and tiny modules A2T are supported (versions 1 - 14).

However, this player doesn't replace the original a2m.cpp, but picks up all modules that a2m.cpp can't handle.
Ideally, I'd like to remove old a2m.cpp in future since it doesn't support some effects and panning even for versions 1-8.

Now about packers:

```
depack.c
https://ibsensoftware.com/products_aPLib.html
https://ibsensoftware.com/files/aPLib-1.1.1.zip

sixpack.c
https://www.sac.sk/download/pack/ddjcompr.zip
https://github.com/70MM13/Step5Archive/blob/main/resources/sixpack/sixpack.c

unlzh.c
http://fileformats.archiveteam.org/wiki/Ar_(Haruhiko_Okumura)
http://cd.textfiles.com/sourcecode/msdos/arc_lbr/ar002.zip

unlzw.c and unlzss.c
Those unpackers were written in hand-optimized assembly in Adlib Tracker 2, so I did
a backward conversion to C.
It seems that they are customized and don't correspond to any existing C implementation.
```

I've left the names 'depack.c' and 'sixpack.c' because those are original names used by their authors. Note that the old 0.26 algo for aPlib is used which is incompatible with the recent aPlib.
As for 'unlzh.c', it's a combination of multiple original files with some custom tweaking which might make the algo incompatible with the original lzh.
All original copyrights and license details are kept intact along with some notes from the authors.

Yes, the author of Adlib Tracker 2 is notoriously known for using a new pack/unpack routine for every new a2m version...

UPDATE 2024-03-06: Added unlzw.c and unlzwss.c unpackers.


dmitrysmagin:
@mywave82, @Malvineous ?

dmitrysmagin:
@mywave82, @Malvineous ?

mywave82:
Sorry for slow response, been extra busy last months.

CI currently does not compile, so that needs to be resolved.

depack.c: from readme.txt:
```
aPLib is freeware. If you use aPLib in a product, an acknowledgement would be
appreciated, e.g. by adding something like the following to the documentation:

    This product uses the aPLib compression library,
    Copyright (c) 1998-2014 Joergen Ibsen, All Rights Reserved.
    For more information, please visit: http://www.ibsensoftware.com/

You may not redistribute aPLib without all of the files.
```
Probably nice to reach out to the author, since we only use one of the files.

sixpack.c: no special copyright requirement present, published as entry in a magazine competition.

unlzh.c: no special copyright requirement present,

dmitrysmagin:
I've sent a message to contact@ibsensoftware.com asking for permission to use depack.c as-is.
Let's hope the author is checking his mail regularly.

jibsen:
Thanks for taking the time to reach out, you are welcome to use the aPLib decompression code here.

mywave82:
> Thanks! @mywave82 are we ok now? sixpack.c seems to be written 35 years ago and is already used in older player a2p.cpp without much modifications, unlzh.c is public domain too and ~35 years as well, so I don't see problems with merging new a2m-v2.cpp.

Still need to fix the CI that fails.

src/a2m-v2.h
```diff
@@ -0,0 +1,850 @@
...
+
+#include "player.h"
+#include "depack.h"
+#include "sixpack.h"
+#include "unlzh.h"
+#include "unlzss.h"
+#include "unlzw.h"
+
+#ifndef C_ASSERT
+#define C_ASSERT(e) typedef char __C_ASSERT__[(e) ? 1 : -1]
```
Enet4, line 50: As this is C++, you can just use `static_assert`. It results in nicer error messages.

src/a2m-v2.h
```diff
@@ -0,0 +1,850 @@
...
+
+typedef struct {
+    tINSTR_DATA instr_data;
+    uint8_t vibrato;
+    uint8_t arpeggio;
+    tFMREG_TABLE *fmreg;
+    uint32_t dis_fmreg_cols;
+} tINSTR_DATA_EXT;
+
+C_ASSERT(sizeof(tINSTR_DATA_EXT) == 20 + sizeof(tFMREG_TABLE *));
```
Enet4, line 513: This fails in my machine when compiled separately from this project: `sizeof(tINSTR_DATA_EXT)` is 32 instead of 28. I had to instruct this to be packed together without padding with this attribute.

```c++
typedef struct {
    tINSTR_DATA instr_data;
    uint8_t vibrato;
    uint8_t arpeggio;
    tFMREG_TABLE *fmreg;
    uint32_t dis_fmreg_cols;
} __attribute__((packed))  tINSTR_DATA_EXT;
```

Unless there is a compilation flag here that does this automatically for all structs, there may be a need to add this attribute in other structs as well.

Enet4 (commented):
I figured I'd leave some comments on this PR, as I am interested in this A2M loader and I'm seeing that it might get further traction in the near future. Can't speak much about the errors in CI, as I'm not very familiar with the full project.

src/a2m-v2.h
dmitrysmagin: Yep, but static_assert is supported starting from c++11, and we aim to support GCC 4.8 at least. However, we can make the usage of static_assert optional.

Enet4: Should work fine with GCC 4.8 though. https://en.cppreference.com/w/cpp/compiler_support/11

Ultimately we can let CI tell us whether it's ok, assuming that the minimum compiler version is covered.

dmitrysmagin:
Now using static_assert everywhere, and dropped redundant one.
Extra benefit -- static_assert is parsed by code editors and might be highlighted in case of assert failing.

mywave82:
The CI is still failing:
* gcc-4.8 does not have static_assert, that can be fixed with a #define, mentioned earlier in a "resolved" message.
* stresstest fails, as the current code has no checks for buffer-overruns on the loader, this can either be resolved by adding "size" to all places data is parsed
* endian will currently also fail.
* size and endian can be handled more gracefully by using binio more

dmitrysmagin:
- Okay, I'll try to resolve static_assert for gcc 4.8, a good example how to do it is here: https://github.com/wc-duck/dbgtools/blob/master/include/dbgtools/static_assert.h
- Stresstest is not compilable under win/mingw, so I can't say what's exactly wrong
- Endian shouldn't be an issue. A2M/A2T have most data compressed, so binio endian helpers have no sense here. My approach is th efollowing: compressed chunk is read from the file into a memory buffer, then it is unpacked to another memory buffer which is mapped onto a structure which simulates old dos Pascal records (no padding, no alignment, little-endian - all ints/longs are arrays of chars). Then the data is copied field by field to normal structures used by the player which can have padding/alignment and potientially be big-endian.

binarymaster:
binio endian helpers can also operate on memory buffers, iirc.

dmitrysmagin:
Okay, I've understood finally what's wrong with stresstest.
Malformed .a2m files that are rejected by a2m.cpp are not considered bad by a2m-v2.cpp which loads them and tries to play. Thus various kinds of memory corruption occur.
I need to add more validity checks to a2m-v2.cpp

mywave82:
stresstest passes for me locally now, but I am able to find some corner-cases using valgrind. I will check that out later.

Endian I will also look into later.  Most modern machines are little-endian or (bi-endian configured to little-endian), so this is a niche thing. Problem is present when loading 16 and 32bit numbers directly from memory-buffers/file-streams, and is usually easy to fix with no runtime penalty when fileformat matches the machine.

dmitrysmagin:
@mywave82 , thanks for you patch, I was about to make smth very similar. I'll see if it can be improved to look more elegant...

mywave82:
https://ci.appveyor.com/project/AdPlug/adplug/builds/49753291

Appears that MSVC does not like the ... in the switch case blocks

dmitrysmagin:
It's a shame for MSVC, it seems that gcc and clang support '...' in case clauses...
Okay, this might be replaced with multiple case statements in certain places and chained if-else in others.

mywave82:
Looking through the code, and the few parts that probably would be endian sensitive is not in "danger" at the moment, since the memcpy() are commented out.

We are probably ready to merge :-)

dmitrysmagin:
Let's go for it! Better squash commits or leave as is?

barracuda156:
@dmitrysmagin Could you please take a look at these errors? https://github.com/adplug/adplug/issues/247
Quite a number of static asserts in `a2m-v2.h` fail.

PR #212, opened 2024-05-21
Add testing for DOS/DJGPP

AranVink:
This PR is for adding **tests for DOS/DJGPP** to adplug.
In a previous [PR](https://github.com/adplug/adplug/pull/193) DOS/DJGPP support was added.
This builds on top of that, and uses [dosemu2](https://github.com/dosemu2/dosemu2) to run the tests in emulation.

**Details**
Make's test runner is overwritten by calling dosemu instead of trying to execute it as a Linux binary.
Since filenames cannot be longer than 8 characters in DOS, I've had to abbreviate the test programs names to allow them to run in this environment.

When running playertest and stresstest in dosemu, for some weird reason it duplicates the executable names and paths into argv[0] AND argv[1], making the testcode believe it needs to run a specific file. I've added a preprocessor statement to catch this and just execute all the files instead (by setting `argc` to `1`).

**Sidenote**: in the test logs I see playertest throwing warnings about `The player is writing data out of range!` and the stresstest showing `bad refresh: 0`. This is however exactly the same test output as a regular Ubuntu/GCC build as far as I can tell.


mywave82:
Are you able to run the test/strstest.exe manually in dosbox (or possible with a debugger) and detect the failures?

AranVink:
> Are you able to run the test/strstest.exe manually in dosbox (or possible with a debugger) and detect the failures?

Yes, I discovered the test for the compiled version with DJGPP/GCC 4.8 runs flaky. Still investigating why that is.

Side-question: 
Is there any specific reason that the combination Ubuntu/GCC 4.8 is tested?
I've added DJGPP/GCC 4.8 in the past because it is also there in the tests for Ubuntu. If GCC4 is not important I'm all for dropping DJGPP/GCC 4.8 support from Adplug and just stick to DJGPP/GCC12. This saves test time, and [Adplay/DOS for DOS is only built with GCC 12 anyway](https://github.com/adplug/adplay-dos/blob/master/.github/workflows/docker-adplay-build.yml).

mywave82:
> Side-question: Is there any specific reason that the combination Ubuntu/GCC 4.8 is tested? I've added DJGPP/GCC 4.8 in the past because it is also there in the tests for Ubuntu.

It was pulled in as a side-effect of https://github.com/adplug/adplug/pull/176 `Trivia: older mingw/gcc 4.8.x versions don't have std::to_string implementation. And since to_string is used only in one sole case in mid.cpp to show the midi_type value, it can be safely replaced with compatible solution.`

AranVink:
I'm in doubt on how to proceed with the stress test/fuzzing part. I don't have any practical experience in this area, so I hope any of you can give me some direction, or at least verify my train of thought:

### How does current CI with Linux covers this?
For existing CI (Linux) I don't see any automated testing in terms of validating for memory leaks or out of bounds checking. I guess having it execute and fail is the best indication that something is wrong?
My assumption is that any improper handling of memory allocation, any string/input overflow in the code would show up in the Linux version and those tests as well, given the little amount of DOS/DJGPP specific code that's in Adplug.

### Availability of tooling for testing for DJGPP
With DJGPP I **don't** have the `fsanitze=address` or `fsantize=leak` [options](https://github.com/google/sanitizers/wiki/AddressSanitizerLeakSanitizer) available. This pretty much rules out any "standard way" of automated memory testing AFAIK.

I also found a tool called [YAMD](https://www.cs.hmc.edu/~nate/yamd/) for malloc debugging that should work with DJGPP, but unfortunately I couldn't get it to work with the latest version of DJGPP (last updated 1999(!))

DJGPP itself allows for debugging mallocs, but it does not detect any out of bounds writes or other overflow cases by itself. It can check allocation/free errors (see [here](https://www.delorie.com/djgpp/doc/libc/libc_553.html)). You could write a [hook](https://www.delorie.com/djgpp/doc/libc/libc_552.html) into malloc to do some additional tests yourself, but this is way out of my knowledge area.

### My conclusion
Given the above, I think it's a lot of effort and not valuable to try and implement the stress test/fuzzing with DJGPP.
My solution would be just dropping GCC 4.8.5/DJGPP combination (since it was only pulled in as side-effect), and just running GCC 12.2.0/DJGPP stress test and checking the exit code like it's done in the current Linux CI.

Please let me know if you agree, or think I'm missing anything important here.

mywave82:
CI tests that it compiles, that the output (register writes) are consistent, and that some various stress tests doesn't crash (loading faulty files etc).

Crashes currently does not provide much debug output, that needs to be improved.

Memory leaks and memory abuse is tested manually with valgrind, should be added to CI.

To debug crashes, valgrind and/or gdb needs to be used manually. CI does a bunch of fork/exec calls with arguments telling if it is a subtest or main process I believe.

mywave82:
If you rebase on master, a2m-v2.cpp now includes <math.h>, which should remove the compile error you receive.

AranVink:
I've updated the PR to remove the DJGPP/GCC 4.8 since it's not needed, and also made the stress test output less verbose.

Before:
```text
   bad refresh: 0
   bad refresh: 0
   bad refresh: 0
   ...
```

New:
```text
   bad refresh: 0 0 0 ....
```

Regarding the stress test itself, I did notice there were differences between GCC Linux/DJGPP in terms of output:

- Some fuzzing tests report invalid file on Linux, but give bad refresh/timeouts on DOS
- Differences in play time
- Character encoding/UTF8 handling with dosemu differences, I guess these are to be expected.

I'm not too sure if these issues should be investigated further. My assumption is still that most will be caught on Linux CI in the future, and that for DOS it's simply not worth it in terms of stability/exploitability. If you agree I will clean up this PR and undraft it :)

AranVink:
I've added memory leak/address issues tests for Linux in this PR:
https://github.com/adplug/adplug/pull/217
I think combined with this, and the reasoning on my previous comment this is now ready for review 😄

mywave82:
I believe we are good to go. I will do manual rebase/squash to clean up the commit log. The final diff is not that big

mywave82:
Is "Fix test ordering for DJGPP" needed?

AranVink:
> Is "Fix test ordering for DJGPP" needed?

No it is not, probably a leftover from my name refactoring and testing. It could make sense to execute the emu and crc tests first though, since they can be more of a precheck before the other tests.

AranVink:
Is there anything else I can do to help get this merged?

mywave82:
I have been working on #163 when I had time which also affects github actions, and needed this to brew in my head before merging.

PR #217, opened 2024-06-26
Feature/add memory checks

AranVink:
**Memory leaks and memory addressing issues**
This PR adds CI checks for memory leaks and memory addressing issues. It does this by adding `fsanitize=leak` and `fsanitize=address` to regular Linux GCC builds in CI. Note that for all other OS combinations this is **not** done, so there might be rare cases that errors exists with certain combinations of OSes/compilers.

**Issue in a2m-v2**
When enabling this I found a memory leak in the a2m-v2 format parser, where return is called before calling free on allocated memory.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -73,17 +73,21 @@ jobs:
         #brew install --cask mactex-no-gui
         brew install automake texinfo libtool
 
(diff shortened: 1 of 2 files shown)


mywave82:
I'm going to perform a manual squash and rebase first, to clean up the commit log

PR #219, opened 2024-06-28
Add support for reheadered D00 songs

mistydemeo:
Some old-style D00 songs were reheadered later on. These are identifiable via version numbers with bit 8 set. They can be read like old-style D00 songs by skipping the new header and jumping to the old one instead.

I considered trying to get adplug to read the new-style header, but there's a couple other inconsistencies it wasn't happy with; it seemed easier just to use the old-style header. The Mac OS X EdLibPlay player does the same thing. If you're interested in some samples, the [EdLibPlay download](http://www.sidmusic.org/sidplay/mac/) comes with a number of them. I used Laxity/The_Alibi.d00 as my sample song for this patch.


mywave82:
We need to add a sample song into the test database :-)

mywave82:
@mistydemeo Testfile added looks OK?

mistydemeo:
Yes, that testfile looks perfect - doesn't work in `HEAD`, works with this PR.

PR #223, opened 2024-07-22
Revert a minor defect introduced by e1782f426643e92d219983a973cac6f056c563b6

mywave82:
Files that include herad.h would have AdPlug_LogWrite disabled.

### DIFF
--- a/src/herad.cpp
+++ b/src/herad.cpp
@@ -37,6 +37,13 @@
 #include <stdio.h>
 
 #include "herad.h"
+
+#ifdef DEBUG
+#include "debug.h"
+#else
+#define AdPlug_LogWrite
+#endif
+
 #include "load_helper.h"
 
 const uint8_t CheradPlayer::slot_offset[HERAD_NUM_VOICES] = {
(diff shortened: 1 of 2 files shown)


binarymaster:
Why? There's only one file that includes `herad.h` and it's intended to disable it when no DEBUG is defined, because in this case that function is not available anyway.

Malvineous:
It's generally not a great idea to define such global things like that in header files, because if someone decides to include it somewhere else one day they can get unintended side effects.  This is why things like namespaces were introduced, so you can stop header files from interfering with each other.  Generally speaking, header files should only contain things that other parts of the code will need to use.  If other code doesn't need it, it's best not to put it in the header file at all.

So you're right, in this case it makes no difference either way, but moving those definitions out of the header files will lead to fewer surprises in the future so it's considered best practice to do so.

mywave82:
It is included by adplug.cpp, which is how it was discovered.

binarymaster:
Ah yes, I keep forgetting that. But in this case, maybe make this definition somewhere globally?

PR #225, opened 2024-07-30
Use snprintf() instead of sprintf()

mywave82:
Compilers are starting to complain about sprintf usage.

### DIFF
--- a/src/d00.cpp
+++ b/src/d00.cpp
@@ -509,7 +509,7 @@ void Cd00Player::rewind(int subsong)
 std::string Cd00Player::gettype()
 {
   char	tmpstr[40];
-  snprintf(tmpstr, sizeof (tmpstr), "EdLib packed (version %d)",version > 1 ? header->version : header1->version);
+  snprintf(tmpstr, sizeof (tmpstr), "EdLib packed (version %d)", version > 1 ? header->version : header1->version);
   return std::string(tmpstr);
 }
 
--- a/src/d00.cpp
+++ b/src/d00.cpp
@@ -509,8 +509,7 @@ void Cd00Player::rewind(int subsong)
 std::string Cd00Player::gettype()
 {
   char	tmpstr[40];
-
-  sprintf(tmpstr,"EdLib packed (version %d)",version > 1 ? header->version : header1->version);
+  snprintf(tmpstr, sizeof (tmpstr), "EdLib packed (version %d)",version > 1 ? header->version : header1->version);
   return std::string(tmpstr);
 }
 
--- a/src/d00.cpp
+++ b/src/d00.cpp
@@ -509,7 +509,7 @@ void Cd00Player::rewind(int subsong)
 std::string Cd00Player::gettype()
 {
   char	tmpstr[40];
-  snprintf(tmpstr, sizeof (tmpstr), "EdLib packed (version %d)", version > 1 ? header->version : header1->version);
+  snprintf(tmpstr, sizeof(tmpstr), "EdLib packed (version %d)", version > 1 ? header->version : header1->version);
   return std::string(tmpstr);
 }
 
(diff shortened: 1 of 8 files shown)


src/d00.cpp
```diff
@@ -509,8 +509,7 @@ void Cd00Player::rewind(int subsong)
 std::string Cd00Player::gettype()
 {
   char	tmpstr[40];
-
-  sprintf(tmpstr,"EdLib packed (version %d)",version > 1 ? header->version : header1->version);
+  snprintf(tmpstr, sizeof (tmpstr), "EdLib packed (version %d)",version > 1 ? header->version : header1->version);
```
binarymaster, line 512:
```suggestion
  snprintf(tmpstr, sizeof (tmpstr), "EdLib packed (version %d)", version > 1 ? header->version : header1->version);
```

src/herad.cpp
```diff
@@ -69,9 +69,9 @@ std::string CheradPlayer::gettype()
 {
 	char scomp[12 + 1] = "";
 	if (comp > HERAD_COMP_NONE)
-		sprintf(scomp, ", %s packed", (comp == HERAD_COMP_HSQ ? "HSQ" : "SQX"));
-	char type[40 + 1];
-	sprintf(type, "HERAD System %s (version %d%s)", (AGD ? "AGD" : "SDB"), (v2 ? 2 : 1), scomp);
+		snprintf(scomp, sizeof(scomp), ", %s packed", (comp == HERAD_COMP_HSQ ? "HSQ" : "SQX"));
+	char type[41 + 1];
+	snprintf(type, sizeof (type), "HERAD System %s (version %d%s)", (AGD ? "AGD" : "SDB"), (v2 ? 2 : 1), scomp);
```
binarymaster, line 74:
```suggestion
	snprintf(type, sizeof(type), "HERAD System %s (version %d%s)", (AGD ? "AGD" : "SDB"), (v2 ? 2 : 1), scomp);
```

src/herad.cpp
```diff
@@ -69,9 +69,9 @@ std::string CheradPlayer::gettype()
 {
 	char scomp[12 + 1] = "";
 	if (comp > HERAD_COMP_NONE)
-		sprintf(scomp, ", %s packed", (comp == HERAD_COMP_HSQ ? "HSQ" : "SQX"));
-	char type[40 + 1];
-	sprintf(type, "HERAD System %s (version %d%s)", (AGD ? "AGD" : "SDB"), (v2 ? 2 : 1), scomp);
+		snprintf(scomp, sizeof(scomp), ", %s packed", (comp == HERAD_COMP_HSQ ? "HSQ" : "SQX"));
+	char type[41 + 1];
```
binarymaster, line 73: 40 + 1 is already the size with terminating zero, no need to change that.

src/msc.cpp
```diff
@@ -176,7 +176,7 @@ std::string CmscPlayer::gettype()
 {
   char vstr [40];
 
-  sprintf(vstr, "AdLib MSCplay (version %d)", version);
+  snprintf(vstr, sizeof (vstr), "AdLib MSCplay (version %d)", version);
```
binarymaster, line 179:
```suggestion
  snprintf(vstr, sizeof(vstr), "AdLib MSCplay (version %d)", version);
```

src/herad.cpp
mywave82:
```
 1234567890123456789012345678901234567890
"HERAD System AGD (version 2, HSQ packed)"
```
Yes you are right, 40 + 1 is enough.

mywave82:
It is a bit messy @binarymaster that the different files have different coding styles regarding indenting, placement of space etc.

binarymaster:
Yeah it's true, maybe we would need to apply standard formatting at some point.

PR #227, opened 2024-07-31
[BMF] Fix two issues reported by coverity:

mywave82:
* bmf.title[] and bmf.author[] could lack zero-termination on bad input.
* On x86-64, the bitshift operation would result in the upper 32bit being sign-extended as the bitshift defaulted to signed int.

Using `uint32_t` might be more consistent than `unsigned long`

### DIFF
--- a/src/bmf.cpp
+++ b/src/bmf.cpp
@@ -200,10 +200,10 @@ bool CxadbmfPlayer::xadplayer_load()
   if (bmf.version > BMF0_9B) {
     if (tune_size - ptr < 4)
       return false;
-    unsigned long sflags = (tune[ptr] << 24)
-                         | (tune[ptr+1] << 16)
-                         | (tune[ptr+2] << 8)
-                         | tune[ptr+3];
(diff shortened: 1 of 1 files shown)


src/bmf.cpp
```diff
@@ -147,8 +149,10 @@ bool CxadbmfPlayer::xadplayer_load()
   if (bmf.version > BMF0_9B) {
     if (tune_size - ptr < 4)
       return false;
-    unsigned long iflags = (tune[ptr] << 24) | (tune[ptr+1] << 16)
-                         | (tune[ptr+2] << 8) | tune[ptr+3];
+    unsigned long iflags = (((unsigned long)tune[ptr]  ) << 24) |
+                           (((unsigned long)tune[ptr+1]) << 16) |
+                           (((unsigned long)tune[ptr+2]) <<  8) |
+                            ((unsigned long)tune[ptr+3]);
```
binarymaster, line 155: There's similar patter with `sflags` on lines 203-206.

PR #228, opened 2024-07-31
A2M v2 guard (triggered ASAN failure in stresstest) + Remove sixpack.c, it is already implemented in a2m.cpp, which also contains some minor bugfixes.

mywave82:


dmitrysmagin:
My initial thought was to exactly move all pack/depack routines into separate files and make a2m.cpp and a2m-v2.cpp reference them. And eventually remove a2m.cpp in favor of a2m-v2.cpp which should be more accurate.
This commit seems like a weird step back.

mywave82:
The initial reason for the change was that Coverity detected unreachable code, which was fixed in the copy inside a2m.cpp. Also that copy had received some love regarding adding more enums, not using global/static memory and not exposing a symbol that is very generic in its name.

But I do agree that moving the sixpack (sixdepak) code into its own file again is good idea.

There are also minor issues in a2m-v2.cpp detected by Coverity that you could take a look at - I have sent you an invitation, and can resend if needed.

dmitrysmagin:
I see no problem moving sixdepack class into sixpack.cpp if you want to keep it as a c++. The reason I kept plain c was to keep the source as close to original as possible.
Thanks for Coverity invite, I'll examine all warning it shows.

PR #230, opened 2024-08-04
[PSI] ensure that le16() works as expected on all architectures (detected by Coverity)

mywave82:
(waiting for a new scan to complete)


mywave82:
Coverity is too aggressive here.

PR #232, opened 2024-08-05
[IMF] Use correct delete syntax, and remove warnings seen in Appveyor Visual Studio 12 build.

mywave82:


src/imf.cpp
```diff
@@ -59,7 +59,7 @@ bool CimfPlayer::load(const std::string &filename, const CFileProvider &fp)
   if (!f) return false;
 
   // file validation section
-  size_t hdr_size = 0;
+  unsigned long hdr_size = 0;
```
binarymaster, line 62: What's wrong with using `size_t`? I always thought this unit is used for such cases

binarymaster: If it's mixed with `unsigned long` in other places, it's better to fix those.

mywave82: size_t is either a int or long, depending on compiler environment and defines before including lib-c header files, and is mostly only used by libc in the fileoperations; libbinio uses unsigned long, so no need to use size_t anyway in this use-case.

binarymaster (approved):

PR #235, opened 2024-08-26
[A2M v2] Fix some warnings given by Coverity

dmitrysmagin:
Coverity shows a lot of warnings for a2m-v2.cpp (possible overruns etc), will try to handle them here.

Besides that, noted that macro vibrato was being incorrectly applied (and one of the Coverity warnings indicated a suspicious if-else clause that was always true).
Fixed that. This effect is most prominent in fm63b_rv.a2m tune.


src/a2m-v2.cpp
```diff
@@ -747,7 +751,7 @@ void Ca2mv2Player::set_ins_volume(uint8_t modulator, uint8_t carrier, int chan)
     if (modulator != BYTE_NULL) {
         uint8_t regm;
         bool is_perc_chan = instr->fm.connect ||
-                            (percussion_mode && (chan >= 16 && chan <= 19)); // in [17..20]
+                            (percussion_mode && chan >= 16); // in [17..20]
```
Malvineous, line 754: This changes the logic for the case `chan == 20`, just confirming that's the intention?  There are five percussion channels on the OPL chip so I presume it's correct, but then the comment `17..20` no longer holds, it should read `17..21`.

src/a2m-v2.cpp
```diff
@@ -2596,6 +2602,12 @@ void Ca2mv2Player::update_extra_fine_effects()
     }
 }
 
+void Ca2mv2Player::set_current_order(uint8_t new_order)
+{
+    assert(new_order < 0x80);
```
Malvineous, line 2607: Should this be an assert?  If we play a corrupted file then line 2676 might pass an invalid value here, and a corrupted file should never trigger an assertion failure.  It would be better to perhaps print a warning but handle it gracefully as you do on the following line.

Malvineous (approved):

src/a2m-v2.cpp
dmitrysmagin: [17..20] is the remainder from Pascal code which counts array indices from 1 (not 0), so in C terms it should read "in [16..19]

src/a2m-v2.cpp
dmitrysmagin: Yes, in general, all erroneous values should reset the order index to 0. But i suspect, assert() is being removed in production code and has meaning only in debug mode.

src/a2m-v2.cpp
Malvineous: But 17, 18, 19, 20 is four numbers, and you have changed the code from 16, 17, 18, 19 to 16, 17, 18, 19, 20.  So there are now five numbers matching the condition instead of only four.

src/a2m-v2.cpp
Malvineous: Yes but often asserts aren't disabled and it means you won't be able to play a corrupted file if you're running in debug mode.

Asserts are meant to catch bugs in the code, and handling a corrupted file isn't a bug, it's a feature, so in my opinion the assert doesn't belong here because you're not using it to catch a programming error in the code.

To put it another way - if you play a corrupted file and you get an assertion failure, then it should be fixed.  But what code would you change to fix the assertion failure?  None, because there's no bug.

dmitrysmagin: Okay, replaced asserts in set_current_order() and set_ins_volume() to log warnings. A agree that we should try to ignore as much errors as possible when trying corrupt file.

src/a2m-v2.cpp
dmitrysmagin: At the beginning of set_ins_volume we do if (chan >= 20) return; so it cuts off the upper bound; later we check for channels >= 16, so it gives 16..19 range;
Seems correct to me.

binarymaster (approved):

PR #238, opened 2024-12-25
[PLX] Coverity reported that fmregs was one byte too small and some attributes that could be used without being initialized.

mywave82:
Also improved formatting of opl2_init_regs[] definition.

Avoid possible division by zero in getrefresh()


mywave82:
Regarding rewind():

Any entries in opl2_init_regs that are zero, will not be set during initialization calls to set_regs due to memset(fmchip, 0, sizeof(fmchip));

Perhaps setregs should had an alternative version that is wright-through no matter what previously is stored in fmchip[], and remove the memset in rewind().

PR #241, opened 2025-02-27
Fix RAD v2 nested riff precedence.

AliceLR:
In RAD v2, when the first line of a riff references another riff, the innermost riff is the riff that continues playing afterward. The upstream RAD v2 replayer source (and thus AdPlug's replayer, which is based on it) has a bug where the innermost riff will be played, but then the outer riff(s) will overwrite the riff track pointer as the recursive call unrolls. The discrepancy between how the tracker and the replayer perform this seems to be due to the tracker playing fully unpacked pattern data (thus it does not need to store an updated track pointer on return).

I've fixed this bug by adding an updated flag to the CRiff struct. When recursively invoking a riff, `TickRiff` will set this flag to `false`, and when a riff is handled, `TickRiff` will set this flag to `true`. If this flag is `true` when `TickRiff` is exiting from a recursive call, it will skip the tasks at the end of `TickRiff` (which are no longer relevant to this clobbered riff). This is the least invasive solution I was able to come up with, and while I can't guarantee it's 100% correct, it has fixed every case I've tried so far.

Adapted from my patch to the MegaZeux fork of the RAD replayer (currently hoping for feedback from Shayde):
https://github.com/AliceLR/megazeux/pull/488

-----

Test file 1: this uses various combinations of nested riffs. Playback should roughly match the WAV recording from the tracker.
[nestriff.rad.zip](https://github.com/user-attachments/files/19000769/nestriff.rad.zip)
[nestriff.wav.zip](https://github.com/user-attachments/files/19000771/nestriff.wav.zip)

Test file 2: this is nearly the same as test 1, but using the transpose effect instead of the regular riff effect.
[nestrifftr.rad.zip](https://github.com/user-attachments/files/19000776/nestrifftr.rad.zip)
[nestrifftr.wav.zip](https://github.com/user-attachments/files/19000777/nestrifftr.wav.zip)

Test file 3: my module where I encountered this bug in practice (because I misunderstood how nested riffs work). The bug occurs in channel 5 from order 8 until order 20. This channel periodically breaks without the patch and plays the same as RAD with the patch. Unfinished module, sorry for any ear pain.
[fn_title.rad.zip](https://github.com/user-attachments/files/19000800/fn_title.rad.zip)
[fn_title_ord8.wav.zip](https://github.com/user-attachments/files/19000801/fn_title_ord8.wav.zip)


mywave82:
(We need to update the CI, so it can complete before handling this PR, and future PRs)

dynamite1981:
> (We need to update the CI, so it can complete before handling this PR, and future PRs)

This should be updated now.

mywave82:
(I rebased your PR to the lastest master, so the CI can execute)

mywave82:
- [ ] Verify that the fix is ok, or we there are other more ideal solutions
- [ ] Add one test file to the CI

mywave82:
I did a quick read on RIFF feature on https://www.3eality.com/productions/reality-adlib-tracker/docs
and recursive RIFF are not mentioned (unless I missed it).

RADPlayer::PlayNote() already have a detection for the corner-case were RIFF end up in a closed self-referenced loop, causing eternal looping and lock-ups.

The CornerCase you want to address is when recursion happens (RIFF calls a RIFF or TRANSPOSE commands), further processing of the current line should stop, as the data is not relevant any-more, and causes overwriting the fresh new pointer to tracker-data.

I agree that your patch fixes the problem regarding recursive usage.

mywave82:
@dmitrysmagin : do you approve this PR?

dmitrysmagin:
@mywave82 
Though, I'm not an author, but yes.
This patch seems to fix certain rad files while not breaking the rest of them, so I consider it useful.

mywave82:
Thank you @AliceLR

PR #245, opened 2025-04-06
Update CI to use current last branch of action/checkout and action/upload-artifact

mywave82:


mywave82:
e491e382479ed2246f124c96c70d94561dce59b9 does not update action/checkout, which is using old version aswell.

binarymaster (approved):

Issue #247, opened 2025-04-12
adplug 2.4 fails to build on static asserts in `a2m-v2.h`: `error: size of array '__STATIC_ASSERT__' is negative` etc. (endianness issue?)


barracuda156:
With gcc-4.2:
```
libtool: compile:  /usr/bin/g++-4.2 -DPACKAGE_NAME=\"adplug\" -DPACKAGE_TARNAME=\"adplug\" -DPACKAGE_VERSION=\"2.4\" "-DPACKAGE_STRING=\"adplug 2.4\"" -DPACKAGE_BUGREPORT=\"\" -DPACKAGE_URL=\"\" -DNDEBUG=1 -DPACKAGE=\"adplug\" -DVERSION=\"2.4\" -DHAVE_STDIO_H=1 -DHAVE_STDLIB_H=1 -DHAVE_STRING_H=1 -DHAVE_INTTYPES_H=1 -DHAVE_STDINT_H=1 -DHAVE_STRINGS_H=1 -DHAVE_SYS_STAT_H=1 -DHAVE_SYS_TYPES_H=1 -DHAVE_UNISTD_H=1 -DSTDC_HEADERS=1 -DHAVE_DLFCN_H=1 -DLT_OBJDIR=\".libs/\" -DWORDS_BIGENDIAN=1 -DHAVE_GETOPT_H=1 -I. -Dstricmp=strcasecmp -I/opt/local/include/libbinio -I./src -isystem/opt/local/include/LegacySupport -I/opt/local/include -pipe -Os -arch ppc -MT src/libadplug_la-a2m-v2.lo -MD -MP -MF src/.deps/libadplug_la-a2m-v2.Tpo -c src/a2m-v2.cpp  -fno-common -DPIC -o src/.libs/libadplug_la-a2m-v2.o
[...]

PR #248, opened 2025-04-17
A2M-V2: Rework loaders and get rid of structures that depend on platform-specific padding/packing

dmitrysmagin:
Spawned by this issue: https://github.com/adplug/adplug/issues/247

What's been done:

- Removed all structures that depend on x86 Little Endian platform and intel-specific packing/padding;
- Now all data is read byte-by-byte and all data is initialized field-by-field (no obscure memcpy for structures)
- Totally separated data structures for the player and for importing data, so the player doesn't rely on x86 specific structures
- STATIC_ASSERT is retained for two cases - to check if byte(char) arrays are not padded (maybe too paranoid...)

As a result, the code must become compilable and workable on PowerPC and Big-Endian archs, though I cannot test it fully. 
@mywave82 , please review.


mywave82:
gdb is not part the gcc/g++/binutils that is stock available for cross-compiling in debian/ubuntu

mywave82:
When load is about to exit, I dump some information:

songinfo->patt_len=0
songinfo->nm_tracks=0
songinfo->tempo=0
songinfo->speed=0
songinfo->macro_speedup=0
play_line current_pattern=0, current_line=0


So that looks wrong. And play_line is only called once, and it does nothing due to nm_tracks being zero.

dmitrysmagin:
Okay, then need to debug a2m_read_varheader() and then a2m_read_songdata() which fills all the info on number of patterns, speed and tempo.
The first requires that these asserts are not hit while compiling and they are not:
```
// Make sure compiler doesn't pad byte arrays
STATIC_ASSERT(sizeof(uint8_t[2]) == 2, "sizeof(uint8_t[2]) != 2");
STATIC_ASSERT(sizeof(uint8_t[4]) == 4, "sizeof(uint8_t[4]) != 4");
```

mywave82:
Problem is likely in a2t_depack().
A2M_SONGDATA_V9_14_NM_TRACKS() returns zero, even if I move the code up to just after it is called (I'm inside a2m_read_songdata)

(just verified that a2m_read_varheader() is generating the same len[] table content, so code behaves the same on x86_64 and ppc until we call a2t_depack)

dmitrysmagin:
This is obviously NOT big-endian friendly:
image

We should imply that the encoded data is LE, but the code could be both BE and LE.

dmitrysmagin:
@mywave82 
Fixed LZH_decompress(), should work now. Could you try again?

mywave82:
If you want, you can clean up the commit history / rebase / squash some of the commits.

dmitrysmagin:
Yep, squashed commits logically into two: 1) a2m/a2t loader rework and 2) Fix for note-off+portaments effects.
Now can be merged.

mywave82:
@dmitrysmagin : is PR ready for merge?

dmitrysmagin:
Yes, please.

PR #255, opened 2026-04-13
CMake installs includes.

RobertP-McDowell:
Sorry for such a small pr. Simply installs headers files from CMake builds, to `adplug/out/install/.../include/adplug/` for example, just as libbinio does. Makes compiling with this library slightly quicker.


binarymaster:
Thanks!

PR #256, opened 2026-05-08
Stresstest omit on windows platform (configure/automake)

dmitrysmagin:
Trivia:
test/stresstest.cpp is very unix-specific and is not directly portable to Windows.
Since it is already omitted in cmake files and msvs project files, it is natural to ignore it on windows platform.

What this fix does:
Eliminates stresstest.cpp from building on Windows platform (mingw/cygwin variants), so 'make check' passes without problems

PR #257, opened 2026-05-09
stresstest: Properly port to Windows, make compilable by mingw**

dmitrysmagin:
Here's the opposite thing made in https://github.com/adplug/adplug/pull/256
Instead of omitting stresstest, make it compilable under Windows.

Key changes:
* Unix-specific headers isolated with conditional ifdef/endif
* alarm() replaced with CreateThread() based emulation
* stat() is replaced with _stat(), access() with _access(), 
* system() returns different statuses than unix - handled
* s.append(s, pos); is replaced with s.append(s, pos, std::string::npos); - mingw 4.8.0 specific fix

### DIFF
--- a/test/stresstest.cpp
+++ b/test/stresstest.cpp
@@ -27,16 +27,33 @@
 #include <cstring>
 #include <string>
 #include <signal.h>
+#ifdef _WIN32
+#include <windows.h>
+#include <io.h>
+#include <sys/stat.h>
+#else
 #include <unistd.h>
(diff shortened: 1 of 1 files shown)


mywave82:
Before merging, thinking if we should add a mingw/wine to the CI, to make it complete

dmitrysmagin:
@mywave82
Could you, please, merge this?
I've tested this modification on Windows, seems to work reliably.
We can create a separate issue for CI with mingw/wine.

PR #259, opened 2026-05-28
VGZ: Add support for gzipped .vgm files

dmitrysmagin:
Hi.

Recently I've noticed that adplug doesn't support gzipped .vgm files, while https://opl.wafflenet.com/ site keeps the entire collection exclusively in .vgz form... Not very handy

What's done:

- Modified existing vgm.cpp to load the entire file into the buffer first
- This is needed, because libbinio can't read from memory buffers, only from files (at least, without modifications)
- Simple ungzip.c added to unpack the file if it's .vgz
- Further on the file is treated as regular .vgm
- Added a couple of vgz into test, all pass


Malvineous:
Looks like the gunzip code is public domain (although it only seems to mention the original was public domain) so hopefully that means we're ok as far as licensing goes.

binarymaster (approved):

PR #260, opened 2026-05-28
VGM: Add support for OPL1 (YM3526) file type

dmitrysmagin:
Hi again.
After trying some files from https://opl.wafflenet.com/ I've noticed that some rare ones are marked as OPL1. After some quick research it turned out that it's YM3526 aka OPL1. It's basically the same as its successor OPL2 with the exception of having no waveform select bank (all operators are sine waves). The rest of the register set is the same.

So, why not adding support to adplug? We can pretend it's OPL2 with all sine waves. :)

Added one of such tunes to testing suite: RRThiel.vgz
image


binarymaster:
There's a dosemu2 test run problem in CI djgpp build, don't have a solution for it yet. It's unrelated to the changes in this PR.

Opened issue: https://github.com/adplug/adplug/issues/261

dmitrysmagin:
Copilot recommends this:
I'll help you find a solution for this failing job. Let me get the job logs first.## Solution

The test failures are caused by missing KVM virtualization support and the libdj64.so.0 library in the GitHub Actions runner environment. All four tests (playtest.exe, emutest.exe, crctest.exe, and strstest.exe) are failing because they require dosemu2 to run DOS executables, but the environment is missing critical dependencies.
Root Causes:

    KVM permission denied: ERROR: KVM: error opening /dev/kvm: Permission denied
    Missing library: cannot dlopen /tmp/dosemu2_1001/dosemu2_pshm_3uKEbN/COMMAND.COM: libdj64.so.0: cannot open shared object file
    Shell initialization failure: dj64: load failure and Shell E:\\command.com exited with code 0x1

Solutions:

Option 1: Skip DOS tests in CI (Recommended)

Add a conditional check to skip these tests when running in CI or when DOS emulation is unavailable. Update your test configuration or workflow to set an environment variable:
bash

# In your GitHub Actions workflow file (.github/workflows/build.yml)
- name: Build and Test
  env:
    SKIP_DOS_TESTS: 1  # Skip DOS-dependent tests
  run: |
    ./configure && make check

Then modify your test setup to honor SKIP_DOS_TESTS:
Makefile

# In your Makefile or test configuration
if test -z "$SKIP_DOS_TESTS"; then
  # run DOS tests
else
  # skip DOS tests
fi

Option 2: Install missing dependencies

If you want to run the tests, install dosemu2 and required libraries:
bash

- name: Install dosemu2 dependencies
  run: |
    sudo apt-get update
    sudo apt-get install -y dosemu2 libdj64 djgpp

However, this may still fail in GitHub Actions due to KVM restrictions in the container environment.

Option 3: Use a different test runner

Use a runner that supports KVM or use native compilation instead of DOS emulation for testing.
Recommendation:

Implement Option 1 - conditionally skip DOS tests in CI environments since the GitHub Actions runner doesn't support KVM virtualization. Keep these tests for local development or use a self-hosted runner with full virtualization support.

Malvineous:
I've used [EMU2](https://github.com/dmsc/emu2) before for things like this with good success.  Once compiled it can run native DOS .exe files and the output comes back on stdout, so you can include DOS programs as part of build scripts and CI/CD pipelines.  It's fully software emulated so doesn't require hardware KVM support or special permissions to run.

I used it in [testexe](https://github.com/camoto-project/testexe) for compiling C code into a DOS executable using the DOS version of Borland Turbo C, then running DOS utilities like PKLite and LZEXE on the resulting .exe.  The whole process works perfectly fine integrated into a Makefile running on Linux.

AranVink:
There was an upstream package issue on Ubuntu with the dosemu2 packages, should be fixed now.

src/vgm.cpp
```diff
@@ -322,6 +332,8 @@ std::string CvgmPlayer::gettype()
 		strcpy(chip, "OPL3");
 	else if (vgmDual)
 		strcpy(chip, "Dual OPL2");
+	else if (vgmOPL1)
+		strcpy(chip, "OPL1");
```
binarymaster: Ok, let's call it "OPL" as it's being referred to most places.
```suggestion
		strcpy(chip, "OPL");
```

binarymaster (approved):

PR #270, opened 2026-06-29
Add mingw-w64 CI build, fix dir delimeter for win32 in tests/mus.cpp

AranVink:
This PR adds **mingw-w64** to the **CI build**

To get it to work, some tests had to be changed so they use the right path separator (dir delimeter), as well as a minor fix in `mus.cpp`

@dmitrysmagin I noticed you made some recent additions for MinGW as well, hope our work doesn't overlap (was happy to see you fixed some tests which I was in the process of tackling myself as well)

@binarymaster Please double-check the `mus.cpp` modifications if you are able. If I read to code correctly the bank loading first tries it with `/` from the path, then with `\\` if none is found, which to me suggested this was an attempt to split string on path. I guess under most conditions this worked fine so far, but when doing a hybrid `make check` test from Linux, calling the tests under `wine64` this code broke in a weird way. Using a predefined delimiter fixed the issue for mingw (and kept other builds green).

### DIFF
--- a/src/mus.cpp
+++ b/src/mus.cpp
@@ -39,6 +39,12 @@
 #include "debug.h"
 #endif
 
+#if defined(MSDOS) || defined(_WIN32)
+#	define DIR_DELIM	"\\"
+#else
+#	define DIR_DELIM	"/"
+#endif
+
 /*** public methods *************************************/
 
 CPlayer *CmusPlayer::factory(Copl *newopl)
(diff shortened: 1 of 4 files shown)


src/mus.cpp
```diff
@@ -200,7 +204,7 @@ bool CmusPlayer::load(const std::string &filename, const CFileProvider &fp)
 			std::string nam = bankName[nm];
 			std::string ext = "bnk";
 
-			size_t np = filename.find_last_of("/");
+			size_t np = filename.find_last_of(DIR_DELIM);
 			if (np == std::string::npos)
 				np = filename.find_last_of("\\");
```
binarymaster: These should probably be removed now as well:
```suggestion
```

binarymaster (commented):
Yes, in mus.cpp I was attempting to support both slashes for path separation.

src/mus.cpp
AranVink: Woops, missed this one on squash, thanks!

binarymaster (approved):

PR #277, opened 2026-07-18
Let dro player can detect old v0 file

M3MEMonster:
Updated version reading logic to let it can detect old dro file which has no version tag in binary file.

PR #278, opened 2026-07-20
Debug #277 failures

mywave82:
Please ignore this PR, it is for debugging


Malvineous:
Maybe when you're done, it's worth adding something to libbinio to fail on readInt(8) (or even >4) to prevent future accidental use?

mywave82:
I pinpointed the failure for Windows now I think, Testopl:update using printf on floating-point numbers. Lets see how this test-build goes

mywave82:
> Maybe when you're done, it's worth adding something to libbinio to fail on readInt(8) (or even >4) to prevent future accidental use?

Yes, that corner case with undefined-behaviour needs to fixed too.

binarymaster:
> Testopl:update using printf on floating-point numbers

Sounds similar to what @dmitrysmagin found in https://github.com/adplug/adplug/pull/268:

> getrefresh() returns an exact rate 23.125 which is then printed into *.ref via "%.2f" formatter. It turns out that rounding in this case varies between different C++ runtimes. 23.125 would print as 23.12 on the gcc and 23.13 on the msvc (pre-ucrt runtime)

binarymaster:
Yeah it's in fact the same issue. I will revert CMF part in #280, already confirmed that tests are passing.

PR #282, opened 2026-08-09
Added missing .ld0 extension for LOUDNESS Sound System file detection.

M3MEMonster:
Added .ld0 extension for LOUDNESS Sound System in adplug.cpp

### DIFF
--- a/src/adplug.cpp
+++ b/src/adplug.cpp
@@ -130,7 +130,7 @@ const CPlayerDesc CAdPlug::allplayers[] = {
   CPlayerDesc(CxadhypPlayer::factory, "Hypnosis", ".xad\0"),
   CPlayerDesc(CxadpsiPlayer::factory, "PSI", ".xad\0"),
   CPlayerDesc(CxadratPlayer::factory, "rat", ".xad\0"),
-  CPlayerDesc(CldsPlayer::factory, "LOUDNESS Sound System", ".lds\0"),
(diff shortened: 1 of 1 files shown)


mywave82:
Yes, I can see that the support was added in the backend only.

Do you happen to have test-file, so we can include in the test-suite?

binarymaster:
Just want to say, I'm still teasing the idea of dropping *.ld0 extension and dynamically detecting this type of LDS files, that's why I'm not rushing to merge this.

M3MEMonster:
> Yes, I can see that the support was added in the backend only.
> 
> Do you happen to have test-file, so we can include in the test-suite?

There is already a .ld0 file called GENORI.LD0 in the testmus folder

M3MEMonster:
> Just want to say, I'm still teasing the idea of dropping *.ld0 extension and dynamically detecting this type of LDS files, that's why I'm not rushing to merge this.

I think we should keep both methods to handle more scenes.

mywave82:
@binarymaster out in the wild, is the files usually both named .lds, or is .ld0 commonly used. If the latter, I agree with pull-request, and you can always perfect the detection in the loader itself at a later point in time.

binarymaster:
They have .lds extension by default, ".ld0" was added like a hack or workaround: https://github.com/adplug/adplug/commit/79a82894e3a05d3cc84ffbc0090a7f2f49c54190

